PR #15, opened 2018-11-26
add syntax sugar to support JSON array/object value

rosbit:
because awk doesn't support literal JSON object, it is very tedious to create an array variable like this:
```awk
   a["f1"] = 1
   a["f2"] = 2
   a["f3"] = 3
```

I added two syntax sugars to write literal JSON object. The changes includes:
 1.  add a keyword "var" to indicate the following NAME is an array variable name
 2. add some tokens, they are "true", "false", F_JARRAY, F_JOBJECT
 3. F_JARRAY and J_OBJECT are implemented as built-in functions to wrapper arguments to array values
 4. expand value struct in interp package to support value type "map[string]value"

now we can write literal JSON object in awk like the following:
```javascript
  var a = {
       f1: 1, f2: 2, f3:3,
       "a": true, "b": [1, 3, 4]
  }
```

and we can  refer the variable by index like
 `print a["f1"]`

If you think this is useful, merge it to the master trunk. Thank you!

Rosbit Xu.

### DIFF
--- a/interp/functions.go
+++ b/interp/functions.go
@@ -80,6 +80,40 @@ func (p *interp) callBuiltin(op Token, argExprs []Expr) (value, error) {
 			p.setLine(out)
 		}
 		return num(float64(n)), nil
+
+	// for JSON sugar
+	case F_JARRAY:
+		arr := make(map[string]value, len(argExprs))
+		for i, a := range argExprs {
+			arg, err := p.eval(a)
+			if err != nil {
+				return value{}, err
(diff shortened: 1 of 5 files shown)


benhoyt:
Hi @rosbit, this is a nice idea. However, I'd like to keep GoAWK close to `awk` and the POSIX AWK spec, and not add arbitrary features beyond that, so unfortunately I won't be including this.

PR #18, opened 2018-12-18
Avoid type checking for config func and config var

Jeffail:
Type checking was broken (caused a panic) when both the function and the variable used as an argument to a function are defined via config.

This commit simply removes the type check in this particular case. Ideally we would instead perform type checking based on the reflection derived types from the config based functions.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -888,6 +888,41 @@ func TestConfigVarsCorrect(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


benhoyt:
Closing this as I fixed slightly differently and added a couple more tests in https://github.com/benhoyt/goawk/commit/c5f1bc65519297a35f32678fc7edfe80cd04f22e

PR #19, opened 2019-01-19
Add config options for "safe mode" to prevent exec and file reads/writes

benhoyt:
This is useful for running scripts from untrusted user input.

### DIFF
--- a/TODO.txt
+++ b/TODO.txt
@@ -1,5 +1,3 @@
-* should interp have a configurable "safe" mode that doesn't run
-  system(), redirection, etc?
 * support "getline lvalue"
 * fix how we handle numstr type when setting $0 directly, see TODO test:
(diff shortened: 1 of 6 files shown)


benhoyt:
@Jeffail thoughts/comments on this addition? Not sure if benthos would need it as those scripts are probably trusted, but this adds three options to `interp.Config` to disable system()/exec, writing to files, and reading from files.

niemeyer:
That's very nice, thank you. I can easily see this being useful in future integrations of awk-like functionality into other applications.

The other thing that would likely be necessary in this context is some control over execution time, which shouldn't be too hard either.

Jeffail:
Hey @benhoyt, looks great. I wouldn't make immediate use of this but I could definitely see it being a useful feature at some point.

benhoyt:
Thanks for the comments, folks -- merged!

PR #32, opened 2019-09-11
support @n to express AWK ouput field n, which the semantics of $n

lilixiong2018:
There are many cases, $1/$2/... means arg1/arg2/..., for example, Cmder let us to set alias mycp=cp $1 $2 and then use mycp src dst. In this case, $1=src $2=dst

Consider this case, opengiturl=cat .git\config | grep https | goawk "{ print $3 }", here $3 will be set to arg3 of command line "opengiturl arg1 arg2 arg3".  If we try command "opengiturl", $3 will be empty string. But in this case, we expect goawk "{ print $3 }" will print third field output of goawk, which is "https://github.com/lilixiong2018/goawk.git", not arg3.

this commit support @n to express AWK ouput field n, if we let opengiturl=cat .git\config | grep https | goawk "{ print @3 }", and try command "opengiturl", it will print "https://github.com/benhoyt/goawk.git" as expect.

### DIFF
--- a/lexer/lexer.go
+++ b/lexer/lexer.go
@@ -118,7 +118,15 @@ func (l *Lexer) scan() (Position, Token, string) {
 	// through a corpus of real AWK programs to determine actual
 	// frequency.
 	switch ch {
-	case '$':
+	// There are many cases, $1/$2/... means argv1/argv2/...
+	// In order to distinguish between $n and argv[N] in this case,
+	// it will be helpful to support @n to express AWK ouput field n,
+	// which the semantics of $n
+
+	// Example (Windows OS)
+	// echo a = 100 | goawk "{ print @3 }" will print 100, same as
+	// echo a = 100 | goawk "{ print $3 }" will print 100
+	case '$', '@':
 		tok = DOLLAR
 	case '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '.':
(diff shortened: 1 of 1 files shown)


benhoyt:
I'm not sure I understand the motivation for this, and I'd rather not change the AWK lexing. Can you simply escape the `$` in the AWK source with `\$`, something like (untested):

    opengiturl=cat .git\config | grep https | goawk '{ print \$3 }'

lilixiong2018:
Thank you, benhoyt. I try escape the $ in the AWK source with \$, Cmder does not seem to support such escaping.

Here is my user_aliases.cmd setting:
opengiturl=cat .git\config | grep http | for /f %i in  ('goawk "{ print \$3 }"') do explorer %i

THe result of the alias comaand is:
$ opengiturl
----------------------------------------------------------
{ print \ }
         ^
----------------------------------------------------------
parse error at 1:10: expected \n after \ line continuation

lilixiong2018:
I found a solution that didn't require to change lexer. Instead use alias, I write a bat named "opengiturl.bat" with content:

```
cat .git\config | grep http | for /f %%i in  ('goawk "{ print $3 }"') do explorer %%i
```

It seems to work very well

benhoyt:
Sounds good, thanks for the update.

PR #54, opened 2021-05-17
Support wildcard expansion on Windows

zetamatta:
On Windows, since the standard shell CMD.EXE and PowerShell do not expand arguments including wildcards, the some applications have to do instead of shells.

The package [getwild](https://github.com/mattn/getwild) expands wildcards just by importing it.
( The name "getwild" means the Japanese popular song. )

I made the patch using [getwild](https://github.com/mattn/getwild) for goawk . If my patch does not have problems , would you merge ?

Before:

```
$ goawk "FNR==1 { print FILENAME }" *.go
open *.go: The filename, directory name, or volume label syntax is incorrect.
```

After:

```
$ goawk "FNR==1 { print FILENAME }" *.go
goawk.go
goawk_test.go
goawk_windows.go
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,5 @@
 module github.com/benhoyt/goawk
 
 go 1.13
+
+require github.com/mattn/getwild v0.0.1 // indirect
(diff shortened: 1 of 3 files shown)


benhoyt:
Thanks for the PR! I'm good with the idea, however, I don't want to add such a tiny dependency, and in addition, this change needs tests. I'll push a PR shortly that addresses these.

benhoyt:
Thanks again for your contribution. PR #55 with this fixed is merged into master now.

zetamatta:
Thank you for supporting wildcards.
I have understood your judgement that tiny dependency should not be add.
I am satisfied with this feature !

benhoyt:
You're welcome. I also made it so wildcards work with the `-f` flag.

PR #57, opened 2021-07-01
Fix random seed

inkel:
First of all: thank you for this library! It's fantastic.

While working with it I've found that the RNG in `ExecProgram` was always initialised with the same static values, meaning that every time I've run a simple program like `BEGIN { print rand() }` it was printing `0.606997` every single time.

This PR changes the RNG generation to one with a time-based seed, and while at it it also does it by computing the initial seed and random source outside the `ExecProgram`. Doing this still passes the tests that I've added for proving we always get different values, but also reduces the execution time in 80%. I've run the benchmarks 40 times and then used [`benchstat`](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) to compare the results:

```
name            old time/op    new time/op    delta
BuiltinRand-12    10.5µs ± 1%     2.0µs ± 3%  -81.21%  (p=0.000 n=34+36)

name            old alloc/op   new alloc/op   delta
BuiltinRand-12    11.8kB ± 0%     6.4kB ± 0%  -45.69%  (p=0.000 n=40+40)

name            old allocs/op  new allocs/op  delta
BuiltinRand-12      18.0 ± 0%      17.0 ± 0%   -5.56%  (p=0.000 n=40+40)
```

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -192,6 +192,11 @@ type Config struct {
 	NoFileReads  bool
 }
 
+var (
+	randSeed = float64(time.Now().UnixNano())
+	randSrc  = rand.NewSource(int64(math.Float64bits(randSeed)))
+)
+
 // ExecProgram executes the parsed program using the given interpreter
 // config, returning the exit status code of the program. Error is nil
 // on successful execution of the program, even if the program returns
@@ -214,9 +219,8 @@ func ExecProgram(program *Program, config *Config) (int, error) {
 	// Initialize defaults
 	p.regexCache = make(map[string]*regexp.Regexp, 10)
 	p.formatCache = make(map[string]cachedFormat, 10)
-	p.randSeed = float64(time.Now().UnixNano())
-	seed := math.Float64bits(p.randSeed)
-	p.random = rand.New(rand.NewSource(int64(seed)))
+	p.randSeed = randSeed
+	p.random = rand.New(randSrc)
 	p.convertFormat = "%.6g"
 	p.outputFormat = "%.6g"
 	p.fieldSep = " "
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -21,6 +21,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"time"
 	"unicode/utf8"
 
 	. "github.com/benhoyt/goawk/internal/ast"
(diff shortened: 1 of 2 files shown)


benhoyt:
Thanks for this! I'll look more closely in the next few days.

benhoyt:
@inkel Side note: glad you like GoAWK. Out of interest, what are you using it for, or what project are you using it in? I always love hearing about how people are using it.

inkel:
We're adding an AWK pipeline to Grafana Loki as part of an internal hackathon; you can check this PR: https://github.com/grafana/loki/pull/3906

So far it's been working really great! I don't know if it will be added to Loki or not (you can only execute the program for one line at a time, so no `BEGIN` nor `END`).

I might have found another improvement but I need to work out the details yet. I'll probably let you know tomorrow about it, either via an issue or PR of I find the time to code it 😬

benhoyt:
Interesting, thanks!

I took a closer look at your changes, and there are a couple of issues:

* A [`rand.Source`](https://golang.org/pkg/math/rand/#NewSource) is not sharable between multiple goroutines, so I don't think we can reuse a global source for all calls to `ExecProgram`. Users should be able to use ExecProgram safely from multiple goroutines. This also changes the behavior (apart from the fact it's not concurrency-safe): two concurrent interpreters will be using the same stream of random numbers, so each program will see a different stream than if you were running a single program.
* `awk` and `gawk` (and the current `goawk`) both have this behavior where the seed is the same each time. (This is documented with a "caution" in the Gawk docs, see `rand` [here](https://www.gnu.org/software/gawk/manual/html_node/Numeric-Functions.html).) That's why I used the static seed approach originally. You can use `srand()` in your AWK program if you want a different stream each time.

Note that `ExecProgram`, as you've seen, has a pretty high startup cost. It does a whole bunch of allocations and setup, and isn't intended to be called line by line. Is there a way you could call `ExecProgram` once and pass your logs as input to the `interp.Config.Stdin` reader line by line? (You might have to use concurrency or a goroutine to do the sending.) I think this would be a lot more efficient, as it would avoid the `ExecProgram` startup cost each for each log line.

If we did want to add the ability to make the seed different each time (and I'm not opposed to that as a configuration option), I suggest we add a new field to `interp.Config`, something like `Randomize bool`, which if set to true, uses a time-based input as the seed in the `ExecProgram` setup.

inkel:
Hey, thanks for replying so thoroughly! I appreciate that.

I totally understand your points, and in fact I was surprised to learn that indeed a static seed is the standard for all AWK versions.

I'll look into your suggestions, although if implemented then I'll probably do it in another PR. I'm on my phone right now so it's hard to think about the code 😺 

As for the startup cost of `ExecProgram` profiling that was what lead me to this PR, so I'm pretty aware of it. Sadly the way the app where we are implementing GoAWK works this way of having to call that method once per line, although as I'm writing this an idea is coming to mind, so I might get around it. I'll keep you posted!

Thanks again for your time and commitment with this library!

benhoyt:
Cool, thanks! Happy to look at subsequent PRs or comments. In that case I'll close this PR.

PR #61, opened 2021-08-23
See if the Windows example CR vs CR LF newline issue has been fixed

benhoyt:


benhoyt:
Hmmm, this isn't really an issue in `go test`. It's because on Windows, `writeOutput` is doing newline translation, and outputting `\r\n` literally. Just keep skipping these examples on Windows for now.

```go
// Write output string to given writer, producing correct line endings
// on Windows (CR LF)
func writeOutput(w io.Writer, s string) error {
	if crlfNewline {
		// First normalize to \n, then convert all newlines to \r\n
		// (on Windows). NOTE: creating two new strings is almost
		// certainly slow; would be better to create a custom Writer.
		s = strings.Replace(s, "\r\n", "\n", -1)
		s = strings.Replace(s, "\n", "\r\n", -1)
	}
	_, err := io.WriteString(w, s)
	return err
}
```

PR #63, opened 2021-08-28
Add support for getline < "-"

### LINKED ISSUE #62: Add support for `getline < "-"` to read from stdin
gawk, mawk and the "one true awk" have support for this, and it seems reasonable considering that POSIX requires interpretation of `-` as stdin in basically all other areas.

Only issue is that read prompts in general are a bit awkward with #47. For example:
```
$ goawk 'BEGIN { ARGV[ARGC++] = "-"; print "Enter input:"; getline; print "You entered: " $0 }'
this is my input
Enter input:
You entered: this is my input
```

Crestwave:
Resolves #62
Disclaimer: I don't have much experience in Go


benhoyt:
Thanks for this -- I'm happy to add this. A rearranged it a little (returning early led to simpler code IMO), added similar functionality for using `"-"` to print/redirect to stdout as it seemed nice and symmetrical (gawk supports this too, though awk/mawk don't seem to), and added a couple of tests.

Issue #59, opened 2021-08-18
Incorrect `sh` determination for `system()`?


xonixx:
Or at least inconsistent with major implementations:

```
$ PATH= ./soft/gawk51 'BEGIN { system("echo hello") }'
hello
$ PATH= ./soft/bwk 'BEGIN { system("echo hello") }'
hello
$ PATH= ./soft/mawk134 'BEGIN { system("echo hello") }'
hello
$ PATH= ./soft/goawk1.8.1 'BEGIN { system("echo hello") }'
exec: "sh": executable file not found in $PATH
$ ./soft/goawk1.8.1 'BEGIN { system("echo hello") }'
hello
```

AFAIK the `system()` call should use hardcoded `/bin/sh`

benhoyt:
Yep, that looks like right. I've just looked at the Gawk source code and it hardcodes `/bin/sh` and uses `execl()` directly. AWK (bwk) calls `system()` which is documented to use `execl("/bin/sh", "sh", "-c", command, (char *) NULL);` under the hood.

I'd be happy for this to be fixed. Though I think it'd be good to add a field to the `interp.Config` struct called `Shell`, and if it's not set default to `/bin/sh` (on Unix; not sure what a good default is on Windows these days -- cmd.exe, Powershell?).

Are you interested in submitting a PR?

xonixx:
Yes I think I would be able to add a PR.

I propose not to mix the two unrelated issues, though:

1. `/bin/sh` vs `sh` POSIX inconsistency.
2. Handling shell invocation on Windows.

I would be glad to provide a fix for item 1 above. 
To me item 2 requires much deeper consideration, testing, etc. 

@benhoyt Is this OK?

benhoyt:
Yes, that's fine, thanks!

benhoyt:
Squashed and merged your PR in https://github.com/benhoyt/goawk/commit/3e37e2071694d92e0c6ea7e0384339e1c72a4260 -- thanks!

PR #64, opened 2021-09-08
Add ShellCommand config option and change default to /bin/sh (on Unix)

xonixx:
Attempt at fixing #59.

I checked the test suite (`go test ./... -v`) - seems all green.
1. Though I feel would be good to add some more tests for this specific case, but I'm a bit unsure how exactly. The test suite in `interp_test.go` doen't have means for provinding environment variables to test. Would be glad for an advice.
2. Hopefully I changed `sh` -> `/bin/sh` in all relevant places (to cover `system()`, `"cmd" | getline`, `print | "cmd"`). But now it feels like a bit of copy-paste. Do you think it makes sense to do something about this - like extracting a function or a constant? If yes, my concern is - what is better place for it, since the change is in two files, but it doesn't feel right to complicate things much for little reason.

### DIFF
--- a/interp/functions.go
+++ b/interp/functions.go
@@ -192,7 +192,7 @@ func (p *interp) callBuiltin(op Token, argExprs []Expr) (value, error) {
 			return null(), newError("can't call system() due to NoExec")
 		}
 		cmdline := p.toString(args[0])
-		cmd := exec.Command("/bin/sh", "-c", cmdline)
+		cmd := p.execShell(cmdline)
 		cmd.Stdout = p.output
 		cmd.Stderr = p.errorOutput
 		err := cmd.Start()
@@ -261,6 +261,15 @@ func (p *interp) callBuiltin(op Token, argExprs []Expr) (value, error) {
 	}
 }
 
+// executes code using configured system shell
+func (p *interp) execShell(code string) *exec.Cmd {
+	shellInterpreter := p.shellCommand[0]
+	shellArgs := p.shellCommand[1:]
+	shellArgs = append(shellArgs, code)
+	cmd := exec.Command(shellInterpreter, shellArgs...)
+	return cmd
+}
+
 // Call user-defined function with given index and arguments, return
 // its return value (or null value if it doesn't return anything)
 func (p *interp) callUser(index int, args []Expr) (value, error) {
@@ -320,11 +329,11 @@ func (p *interp) callUser(index int, args []Expr) (value, error) {
 	return null(), nil
 }
 
-// Call native-defined function with given name and arguments, return
+// Call native-defined function with given name and arguments,
 // return value (or null value if it doesn't return anything).
 func (p *interp) callNative(index int, args []Expr) (value, error) {
 	f := p.nativeFuncs[index]
-	minIn := len(f.in) // Mininum number of args we should pass
+	minIn := len(f.in) // Minimum number of args we should pass
 	var variadicType reflect.Type
 	if f.isVariadic {
 		variadicType = f.in[len(f.in)-1].Elem()
--- a/interp/functions.go
+++ b/interp/functions.go
@@ -261,12 +261,12 @@ func (p *interp) callBuiltin(op Token, argExprs []Expr) (value, error) {
 	}
 }
 
-// executes code using configured system shell
+// Executes code using configured system shell
 func (p *interp) execShell(code string) *exec.Cmd {
-	shellInterpreter := p.shellCommand[0]
-	shellArgs := p.shellCommand[1:]
-	shellArgs = append(shellArgs, code)
-	cmd := exec.Command(shellInterpreter, shellArgs...)
+	executable := p.shellCommand[0]
+	args := p.shellCommand[1:]
+	args = append(args, code)
+	cmd := exec.Command(executable, args...)
 	return cmd
 }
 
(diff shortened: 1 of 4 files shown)


benhoyt:
Thanks for this. It's a good start, though I think we should make a few improvements:

* As mentioned in #59, while we're at it I think we should make this configurable via `interp.Config`. Let's add a new field `ExecCommand []string`, which if nil, defaults to `[]string{"/bin/sh", "-c"}` on Unix, and `[]string{"sh", "-c"}` on Windows.
* Setting the Windows default to just "sh" without the /bin path will make it keep working as it does today on Windows. See the unit test failures on Windows: https://github.com/benhoyt/goawk/pull/64/checks?check_run_id=3550794167
* I'm open to other names for this config field too. `ExecArgs`?
* With the new field, we will need a couple of additional tests for it. Probably a new `TestExecCommand` function in `interp/interp_test.go`. See for example https://github.com/benhoyt/goawk/blob/0eb3f8c8b7889fe1feff86aed767dcdb30cbc98e/interp/interp_test.go#L1014

xonixx:
First things first I'm totaly agree with your arguments and I'm ready to do the changes you request.

However please find couple of considerations below.

Regarding failing on Windows. 
Just did a quick check https://github.com/xonixx/playground/runs/3568823431?check_suite_focus=true#step:3:8
I see that on Win this uses sh.exe bundled with Git, which is pre-bundled into Windows environment on Github (https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md). 
Therefore I assume on clean windows (no Git) goawk will fail on `system()` as well.

Does it make sense to specify in docs that on Windows the `system()` functionality of goawk is only available when Git or msys (or other sh.exe provider) is installed?

Also does it make sense to not be Windows-specific but rather go more generic approach and implement the adaptive logic to attempt both
- `/bin/sh` as POSIX require
- `sh` from `$PATH` if above fails

@benhoyt if you think it doesn't make sense to go this deep I'm totally fine just to do as you already prosposed.

benhoyt:
It's a good point regarding "sh" only existing on Windows due to Git. However, I'd rather not do a system call in every ExecProgram call due to the cost of hitting the file system (I know some use that in a fairly tight loop). So for now let's do what I suggested, and I can clean up the Windows shell 
handling in a subsequent commit (maybe "cmd.exe").

interp/functions.go
```diff
@@ -261,6 +261,15 @@ func (p *interp) callBuiltin(op Token, argExprs []Expr) (value, error) {
 	}
 }
 
+// executes code using configured system shell
+func (p *interp) execShell(code string) *exec.Cmd {
+	shellInterpreter := p.shellCommand[0]
+	shellArgs := p.shellCommand[1:]
```
benhoyt, line 267: Naming nit: how about we name these `executable` and `args`? We don't need the `shell` prefix in this context as this function already includes that in the name.

interp/functions.go
```diff
@@ -320,11 +329,11 @@ func (p *interp) callUser(index int, args []Expr) (value, error) {
 	return null(), nil
 }
 
-// Call native-defined function with given name and arguments, return
+// Call native-defined function with given name and arguments,
```
benhoyt, line 332: Thanks for catching. Let's make this "return its return value" like the comment on `callUser`.

interp/functions.go
```diff
@@ -320,11 +329,11 @@ func (p *interp) callUser(index int, args []Expr) (value, error) {
...
 // return value (or null value if it doesn't return anything).
 func (p *interp) callNative(index int, args []Expr) (value, error) {
 	f := p.nativeFuncs[index]
-	minIn := len(f.in) // Mininum number of args we should pass
+	minIn := len(f.in) // Minimum number of args we should pass
```
benhoyt, line 336: Good catch!

interp/interp.go
```diff
@@ -247,6 +252,19 @@ func ExecProgram(program *Program, config *Config) (int, error) {
 		}
 	}
 
+	// Setup system shell command
+	if len(config.ShellCommand) != 0 {
+		p.shellCommand = config.ShellCommand
+	} else {
+		var shellInterpreter string
```
benhoyt, line 259: Nits: let's call this `executable`. Also, let's simplify slightly to:

```go
executable := "/bin/sh"
if runtime.GOOS == "windows" {
    executable = "sh"
}
```

interp/interp_test.go
```diff
@@ -1029,6 +1030,25 @@ func TestConfigVarsCorrect(t *testing.T) {
...
+			})
+		testGoAWK(t, `BEGIN { "world" | getline; print }`, "", "hello world\n", "", nil,
+			func(config *interp.Config) {
+				config.ShellCommand = []string{"/bin/echo", "hello"}
+			})
+		testGoAWK(t, `BEGIN { print "hello world" | "-" }`, "", "hello world\n", "", nil,
+			func(config *interp.Config) {
+				config.ShellCommand = []string{"/bin/cat"}
+			})
+	}
```
benhoyt, line 1049: Nice, thanks for adding a test!

For a Windows test, it [looks like](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd) we might be able to test with `config.ShellCommand = []string{"cmd.exe", "/c"}` ... though I'm not quite sure what would be a good, Windows-only command to run. Feel free to leave that out.

benhoyt (changes requested):
Looks good, thanks! Just a couple of style nits.

xonixx:
@benhoyt Hopefully I fixed all CR notes

benhoyt:
Thanks, looks good!

Issue #66, opened 2021-09-24
Parser allows statements directly after an expression


fioriandrea:
The parser allows statements directly after expressions.

Consider this:
```
goawk 'BEGIN{a = 1 print a}'
```

goawk doesn't detect the error, while gawk, mawk, nawk and busybox awk do.

benhoyt:
Hi @fioriandrea -- thanks for the report. Yeah, I'd seen this a while back, but (in my lazyness :-) decided not to worry about it, as it's a superset of valid AWK syntax, so accepting it shouldn't cause problems in valid AWK scripts. However, the parsing does need a bit of cleaning up in this area. I'd be happy if you want to submit a PR?

fioriandrea:
What should be cleaned up? The parsing problem described in the issue title? If that's the case, I can try to fix it by "disallowing" every token corresponding to statements (e.g. print, for, etc.) coming after expressions. I don't know if this is a clean solution though

benhoyt:
I think the problem here doesn't actually have to do with expressions, but with statements being allowed next to each other without a semicolon in between. I've updated the issue description and title to match that. So I think it'll be tightening up / cleaning up the parsing in `stmts`, `stmtsBrace`, or `stmt`. I'm not sure exactly what the details will be, but it'll need to confirm to [POSIX spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html) and obviously all the tests will need to pass. Thanks!

PR #68, opened 2021-09-25
disallow statements immediately after another one

fioriandrea:
This should solve the issue described in [here](https://github.com/benhoyt/goawk/issues/66).
badassign1 fails because the error message is different (I don't think it is an issue). Some tests failed before my changes, so I didn't consider those.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -107,10 +107,11 @@ func (p *Program) String() string {
 // Parser state
 type parser struct {
 	// Lexer instance and current token values
-	lexer *Lexer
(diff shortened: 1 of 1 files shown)


benhoyt:
Thanks! Please go ahead and fix the `badassign1` error message then, so we can see the other tests pass.

I'll look a this further next week. I'm a bit unsure why we need to introduce `prevtok` just for this, when all the other grammar can be done with just `p.tok`-style parsing, but I haven't thought about it too deeply yet.

PR #69, opened 2021-09-25
make number literals behave like awk/gawk

fioriandrea:
Consider this:

```sh
for i in goawk gawk mawk nawk 'busybox awk'; do echo "$i"; $i 'BEGIN{a = 9; e = 19; print 1.2e+a}'; echo ""; done
```

The output is:

```
goawk
------------------------------------
BEGIN{a = 9; e = 19; print 1.2e+a}
                                ^
------------------------------------
parse error at 1:33: expected digits

gawk
1.228

mawk
1.228

nawk
1.228

busybox awk
1.228
```

This means that other version of awk treat things like "1e" like "1 e" (i.e. concatenation). So it is necessary to unread the 'e' if there are no digits after it.


benhoyt:
I added new interp tests to test this behaviour, updated the lexer tests, and factored out `unread` into its own method on `Lexer`. Merging shortly. Thanks for your work on this!

PR #76, opened 2021-11-24
Addresses some IDE warnings relating to GoDoc comments

codefromthecrypt:
This addresses some warnings mostly around the convention of the first
word of a godoc being the same as the exported symbol. This also
addresses some typos and some micronits. Any or all of this can be
dismissed if not welcome. I'm just raising this as noticed while
reviewing the code!

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -456,7 +456,7 @@ func (s *WhileStmt) String() string {
 	return "while (" + trimParens(s.Cond.String()) + ") {\n" + s.Body.String() + "}"
 }
 
-// DoWhileStmt is a fo-while loop.
+// DoWhileStmt is a do-while loop.
 type DoWhileStmt struct {
 	Body Stmts
 	Cond Expr
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -72,7 +72,7 @@ func (e *UserCallExpr) expr()  {}
 func (e *MultiExpr) expr()     {}
 func (e *GetlineExpr) expr()   {}
 
-// Field expression like $0.
+// FieldExpr is an expression like $0.
 type FieldExpr struct {
 	Index Expr
 }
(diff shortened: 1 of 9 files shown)


internal/ast/ast.go
```diff
@@ -184,9 +186,9 @@ const (
 	ScopeLocal
 )
 
-// Variable reference (special var, global, or local). Index is the
-// resolved variable index used by the interpreter; Name is the
-// original name used by String().
+// VarExpr is a variable reference (special var, global, or local).
+// Index is the resolved variable index used by the interpreter; Name
```
codefromthecrypt, line 190: I looked and figure based on common lengths that we are using max 70 columns here.

internal/ast/ast.go
```diff
@@ -232,7 +234,7 @@ func (e *AugAssignExpr) String() string {
 	return e.Left.String() + " " + e.Op.String() + "= " + e.Right.String()
 }
 
-// Increment or decrement expression like x++ or --y.
+// IncrExpr is an increment or decrement expression like x++ or --y.
```
codefromthecrypt, line 237: I made a lot of judgment calls to avoid stuttering where possible, but sometimes it wasn't.

lexer/token.go
```diff
@@ -12,6 +12,7 @@ const (
 	CONCAT // Not really a token, but used as an operator
 
 	// Symbols
+
```
codefromthecrypt, line 15: this extra line prevents the section comment from being mistaken as documentation for `ADD`

benhoyt:
Thanks! I pushed one more tweak, namely changing "etc." to "and so on", as I think it reads better anyway. I'll merge this now.

benhoyt:
Out of interest, are you using GoAWK for any project in particular, or did you just stumble across it?

codefromthecrypt:
Thanks for merging!

> Out of interest, are you using GoAWK for any project in particular, or did you just stumble across it?

I'm studying your work because we are thinking about a parser for WebAssembly's text format, which has some interesting twists and turns in it. I also have been looking at mugo for the same reason. Being able to parse `wat` allows some interesting use cases including helping those learn about WebAssembly and also making conversion easier to use (ex `go run X` vs installing a compiled system package) https://github.com/tetratelabs/wazero/issues/59

benhoyt:
Nice, thanks for the info!

Issue #35, opened 2020-03-15
Reconsider handling of bytes vs characters in length() and similar functions


benhoyt:
Per @arnoldrobbins comment https://github.com/rmyorston/busybox-w32/issues/174#issuecomment-599194834, the POSIX spec says that `length` and similar functions should work with characters, not bytes. The [POSIX spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html) says:

> The index, length, match, and substr functions should not be confused with similar functions in the ISO C standard; the awk versions deal with characters, while the ISO C standard deals with bytes.

Given that, I wonder why (onetrueawk) `awk`'s `length()` function counts bytes instead of characters, at least by default. Here's output on my macOS Catalina machine:

```
$ echo 絵 | nawk '{print length}'    # onetrueawk/awk 20200228
3
$ echo 絵 | gawk '{print length}'    # gawk 5.0.1
1
$ echo 絵 | mawk '{print length}'    # mawk 1.3.4 2020012
3
$ echo 絵 | goawk '{print length}'    # goawk 1.6.0
3

# for reference:
$ locale
LANG="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_CTYPE="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_ALL=
```

So I feel good that GoAWK has the majority on its side :-), but as you point out, the `nawk/mawk/goawk/ behavior does seem wrong according to POSIX. What do you recommend, @arnoldrobbins?

arnoldrobbins:
You ask:
> I wonder why (onetrueawk) awk's length() function counts bytes instead of characters

The answer is that onetrueawk is not POSIX compliant, at least in this regard. Check out the last entry in the BUGS section of its man page, which states that it only supports eight-bit characters. A similar consideration applies to mawk.

I recommend that goawk be POSIX compliant and work with characters, not bytes. In the worst case, you may also wish to adopt gawk's -b option, although I don't know that I actively recommend that.

Hope this helps.

benhoyt:
[...]

PR #82, opened 2021-12-29
Get length of characters (runes) not bytes

msingle:
Referenced in #35 
Maybe it should be configurable?

### DIFF
--- a/interp/functions.go
+++ b/interp/functions.go
@@ -100,9 +100,9 @@ func (p *interp) callBuiltin(op Token, argExprs []Expr) (value, error) {
 	case F_LENGTH:
 		switch len(args) {
 		case 0:
-			return num(float64(len(p.line))), nil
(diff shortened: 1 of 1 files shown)


benhoyt:
Hi @msingle, thanks for the contribution. However, I'm not going to merge this for a couple of reasons: 1) it only addresses the `length()` function, not the other functions like `substr` which should also change to use runes instead of bytes, and 2) it'd be better to use [`utf8.RuneCountInString()`](https://pkg.go.dev/unicode/utf8#RuneCountInString) to avoid the allocation and copying to a slice of runes.

I've been thinking a bit about #35 recently, and I'd like to consider it more holistically there.

benhoyt:
FYI, this was addressed in https://github.com/benhoyt/goawk/pull/83 and is in v1.11.0 of GoAWK.

PR #83, opened 2022-01-02
Use char instead of byte indexes for length, substr, index, and match

### LINKED ISSUE #35: Reconsider handling of bytes vs characters in length() and similar functions

benhoyt:
I've done this as a runtime config setting, Config.Bytes, and the
default is character mode. So this is a backwards-incompatible change,
but I think it's 1) warranted given GoAWK tries to conform to POSIX,
and 2) won't break most scripts, even ones that use non-ASCII, unless
they use constant indexes for substr() on non-ASCII strings.

This definitely affects performance, as some operations that were O(1)
are now O(N) in the length of the string. Here are the results for the
relevant functions (though obviously it'll depend on the string len):

```
name             old time/op  new time/op  delta
BuiltinLength-8   615ns ± 3%  4179ns ± 3%  +579.68%  (p=0.008 n=5+5)
BuiltinSubstr-8   985ns ± 0%  3874ns ± 0%  +293.12%  (p=0.008 n=5+5)
BuiltinIndex-8   1.24µs ± 0%  4.70µs ± 6%  +278.85%  (p=0.008 n=5+5)
BuiltinMatch-8   2.90µs ± 0%  3.30µs ± 3%   +13.70%  (p=0.008 n=5+5)
[Geo mean]       2.35µs       3.98µs        +69.30%
```

Fixes #35


benhoyt:
FWIW, the "real world" benchmark comparison is as follows (`orig` is the master branch, v1.10.0, and `goawk` is this branch):

```
Test      | goawk |  orig |   awk 
--------- | ----- | ----- | -----
tt.01     |  3.53 |  2.65 |  1.00
tt.02     |  1.85 |  1.82 |  1.00
tt.02a    |  1.81 |  1.90 |  1.00
tt.03     |  1.16 |  1.36 |  1.00
tt.03a    |  1.42 |  1.34 |  1.00
tt.04     |  0.84 |  0.85 |  1.00
tt.05     |  0.98 |  0.96 |  1.00
tt.06     |  1.27 |  1.44 |  1.00
tt.07     |  1.46 |  1.45 |  1.00
tt.08     |  1.85 |  2.59 |  1.00
tt.09     |  2.33 |  2.29 |  1.00
tt.10     |  0.39 |  0.40 |  1.00
tt.10a    |  0.37 |  0.36 |  1.00
tt.11     |  2.15 |  2.38 |  1.00
tt.12     |  1.41 |  1.42 |  1.00
tt.13     |  2.69 |  2.63 |  1.00
tt.13a    |  1.92 |  1.92 |  1.00
tt.14     |  1.18 |  1.18 |  1.00
tt.15     |  0.61 |  0.65 |  1.00
tt.16     |  1.27 |  1.30 |  1.00
tt.big    |  1.62 |  1.61 |  1.00
tt.x1     |  1.53 |  1.56 |  1.00
tt.x2     |  0.99 |  0.98 |  1.00
--------- | ----- | ----- | -----
**Geo mean**  | **1.43** | **1.43** | **1.00**
```

So overall I'm pretty happy with how this turned out. Thanks @arnoldrobbins for pushing on this. :-)

PR #87, opened 2022-01-04
Bytecode test

benhoyt:
Comparing against current master (65135b4d173e52573462a229a8f0083fd12729c2), the performance comparison is as follows:

* A simple count-and-sum loop compares, i.e., the script `BEGIN { for (i=0; i<100000000; i++) s += i; print s }` runs in 7.4s on master, 2.8s on bytecode-test (2.6x as fast).
* A more "real world" count-words script `{ for (i = 1; i <= NF; i++) counts[tolower($i)]++ }` runs in 2.5s on master, 1.8s on bytecode-test (1.4x as fast).

This to me says it's worth it -- even the 1.4x as fast or 40% faster would be a nice improvement. It's a fair bit of work to convert, though!


benhoyt:
Closing this in favour of the much more complete `bytecode` branch.

PR #89, opened 2022-01-22
Try making execute call opcode functions from a slice of func

benhoyt:
Surprisingly (to me), it seems somewhat faster! Here is the before and after (these results are consistent across multiple runs):

```
$ time perflock ./goawk_switch 'BEGIN { for (; i<10000000; i++) s+=i+i+i+i+i+i+i+i+i+i }'
real    0m1.944s
...
$ time perflock ./goawk_funcslice 'BEGIN { for (; i<10000000; i++) s+=i+i+i+i+i+i+i+i+i+i }'
real    0m1.682s
...
```

The only difference is that not all instructions are copied to functions yet. I'll try that too, though according to my mental model it shouldn't change the results (low-level perf is funny, though).

I may also switch to panic-based error handling to avoid an error return everywhere (that improved speed a bit too). It would of course be recover()ed at the top-level ExecProgram.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -114,13 +114,14 @@ type interp struct {
 	matchStart       int
 
 	// Misc pieces of state
-	program     *parser.Program
-	random      *rand.Rand
-	randSeed    float64
-	exitStatus  int
-	regexCache  map[string]*regexp.Regexp
-	formatCache map[string]cachedFormat
-	bytes       bool
+	program      *parser.Program
+	compiledProg *compiler.Program
+	random       *rand.Rand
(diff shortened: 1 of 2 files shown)


benhoyt:
Hmm, once all is said and done this isn't as good as I thought -- maybe a 2% speed increase, and messier code (IMO). Here's a branch with the full "funcs" approach in place: https://github.com/benhoyt/goawk/compare/bytecode...bytecode-funcs

Plus, maybe in Go 1.19 we'll get faster switches via jump tables, see:
https://go-review.googlesource.com/c/go/+/357330/

PR #95, opened 2022-01-24
Revert using char instead of byte indexes for length/substr/index/match

### LINKED ISSUE #93: Major speed regression for gron.awk in goawk 1.11.0+
I know that this is sort of expected as it changes the logic of string-related functions to be more POSIX-compliant. 
And I can confirm that adding `-b` flag makes it as fast as before. 

BUT! What is very interesting, GAWK is able to show very close (I would say, the same) speed for with and without `-b` switch!
Did they find a way to make string functions O(1), not O(N)? 
If yes, shell we use the same approach for GoAWK?

If need be I can provide more info and instructions to reproduce the results I've found.

```
=====================
Benchmarking gawk51 -b ...
24571

real    0m1.150s
user    0m1.148s
sys     0m0.008s

=====================
Benchmarking gawk51 ...
24571

real    0m1.179s
user    0m1.182s
sys     0m0.000s

=====================
Benchmarking mawk134 ...
24571

real    0m0.704s
user    0m0.711s
sys     0m0.000s

=====================
Benchmarking goawk1.13.0 -b ...
24571

real    0m1.597s
user    0m1.882s
sys     0m0.077s

=====================
Benchmarking goawk1.13.0 ...
(HANGS...)
```

The command I use for benchmark is 
```
time $AWK -f gron.awk test_data/big.json | wc -l
```

The files gron.awk and test_data/big.json can be found in https://github.com/xonixx/gron.awk repo.

benhoyt:
The reason is because the new O(N) behavior of these functions was just too problematic, and caused "accidentally quadratic" issues such as: https://github.com/benhoyt/goawk/issues/93

This reverts commit b7ec795b6716aa2159907cce7a54351dc78b0788, but leaves in the new bytes tests that were added.

Unfortunately these means *another* small backwards compatibility change, but hopefully not too many people were on the new version and using `interp.Config.Bytes` or `goawk -b` yet, as it's pretty recent. So I think it's worth doing to fix the performance killer.

Later we can figure out a better way to add it back in without the O(1) -> O(N) performance regression.

Fixes #93


benhoyt:
I'm going to go ahead and merge, and reopen https://github.com/benhoyt/goawk/issues/35

Issue #94, opened 2022-01-23
Add API to reduce per-execute allocations


benhoyt:
Per @jfesler's comments on https://github.com/benhoyt/goawk/issues/91, we'd like to add a new API to allow callers to reuse the relatively expensive initialization (mostly due to allocations) in `ExecProgram` when running the same program over and over again on different inputs. Probably something like this:

```
// do this once
program, err := parser.ParseProgram(src, parserConfig)
if err != nil { /* handle error */ }
interpreter, err := interp.New(program) // would return a new exported interp.Interp
if err != nil { /* handle error */ }

// do this as many times as necessary
status, err := interpreter.Execute(interpConfig)
if err != nil { /* handle error */ }
```

benhoyt:
This new API was added in https://github.com/benhoyt/goawk/pull/100

PR #100, opened 2022-02-06
Add API to reduce allocations when re-executing the same program

benhoyt:
As described in https://github.com/benhoyt/goawk/issues/94, for some use cases it'd be good to have a new API to allow reusing the heavy allocations/initialization that `ExecProgram` does to set up the interpreter. You use it like so:

```go
// do this once to parse the program and set up
prog, err := parser.ParseProgram([]byte(`BEGIN {}`), nil)
// handle err
p, err := interp.New(prog)
// handle err

// do this many times
p.ResetVars() // this is optional (call if you want to reset the variables between runs)
status, err := p.Execute(&interp.Config{
	Stdin:   inputReader,
	Output:  ioutil.Discard,
	Environ: []string{},
})
// handle err
```

Benchmarks of the existing `ExecProgram` API against this new one show just how much the overhead is reduced when repeating the execution of a program. For a trivial program, it's 100x (!) as fast when you don't do I/O, and almost 20x as fast when you do. The memory allocations per execute are also vastly reduced.

```
$ go test ./interp -run=^$ -bench=Repeat -benchmem
goos: linux
goarch: amd64
pkg: github.com/benhoyt/goawk/interp
cpu: Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
BenchmarkRepeatExecProgram-8     	   82119	     13686 ns/op	   12680 B/op	      17 allocs/op
BenchmarkRepeatNew-8             	 9028881	       126.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkRepeatIOExecProgram-8   	   51138	     22887 ns/op	   78488 B/op	      30 allocs/op
BenchmarkRepeatIONew-8           	  976191	      1224 ns/op	     240 B/op	      11 allocs/op
PASS
```


jfesler:
I like it. It’s also a lot cleaner looking than my last stab at this, particularly due to breaking those up.

jfesler:
Processing 750k records. simple single runs.   (It's what I can bang together before bed)

|version|cpu time, seconds|wall time, seconds|
|----|----|----|
|go-awk 1.7.0 stock | 466.0 | 75.15 |
|go-awk 1.7.0 forked| 169.8 | 25.8 |
|go-awk 1.15.0 stock| 457.7 | 73.7 |
|**go-awk 1.15.0 with reuse** :star:| 35.0 | 10.6 |
| without awk (baseline) | 4.8 | 3.6 |

Your results are looking *much* better than the hack job I did.

jfesler:
And for those watching: I'm running the script once *per record* (my input data is not traditional input; the usual awk methods of record separation don't apply).

benhoyt:
Thanks, @jfesler, those results are very helpful ... and look great for your use case! One more thing: is the fact that interpreter state (values of variables, arrays, and so on) actually what you want for your use case? Or do you want variables to continue from where they left off between each `Execute` call?

We could design the API a bit differently to require a manual `interpreter.Reset()` call between each run. And if you want to keep the values of variables between runs, you don't call `Reset`.

jfesler:
Exposing a Reset won’t hurt my case, and may even help cut down a few more allocations.

benhoyt:
Updated: added an explicit `ResetVars` method, so you need to call that if you want to reset variables to their (null) defaults between `Execute`s. It's a little more efficient if you don't need to all it, plus it gives people flexibility if they *do* want to keep the values of variables/arrays from the previous run.

jfesler:
- added .Reset
 - added explicitly cached os.Environ()

I found that by not passing os.Environ myself, 40% of the allocations were coming from go-awk doing os.Environ().  Easy enough to cache and pass that in.

|version|cpu time, seconds|wall time, seconds|
|----|----|----|
|go-awk 1.7.0 stock | 466.0 | 75.15 |
|go-awk 1.7.0 forked| 169.8 | 25.8 |
|go-awk 1.15.0 stock| 457.7 | 73.7 |
|go-awk 1.15.0 with reuse| 35.0 | 10.6 |
|go-awk 1.15.0 with latest PR, without Reset() | 30.6 | 7.9 |
|go-awk 1.15.0 with  latest PR, with Reset | 30.6 | 7.9 |
|go-awk 1.15.0 with  latest PR, with Reset, and cached os.Environ :star: | 18.8 | 6.2 |
| without awk (baseline) | 4.8 | 3.6 |


This is pretty amazing.  466 CPU seconds -> 18.8 CPU seconds.

benhoyt:
Those results are awesome, thanks for posting!

Ah yes, I was setting `config.Environ` to `[]string{}` in the benchmarks. I've just pushed a commit which documents how that fields works a bit more clearly, especially in the context of `New...Execute`.

I think I'm going to merge this now. I won't release a new version just yet, so please let me know here or in a new issue if you find any problems with it.

PR #123, opened 2022-05-22
Resolve types using topological sort to avoid O(N^2) worst case

### LINKED ISSUE #98: Resolve types using topological sort to avoid O(N^2) worst case
In `parser.resolveVars` we currently loop through functions to resolve their parameter types in Go's map iteration order (which is randomized), so if you have a whole bunch of functions with unknown types (e.g., functions that just pass their parameter to another function), only a few will get resolved each iteration, and we'll have to do many iterations. Each iteration is O(N) in the number of functions, so the overall process is O(N^2) in the worst case. This is uncommon for real AWK code, but very bad for pathological cases.

There was an iteration maximum that was 10,000. But for the pathological case of thousands of such functions that took ages (over a minute, and I didn't wait any longer). So in https://github.com/benhoyt/goawk/commit/afcb42a9f8889c2a38da7f138d673fd4da6e6a08 I reduced the max iterations to 500, which takes 100ms or so to resolve. It would be highly unlikely for AWK scripts to have 500 functions with unresolved types.

In any case, we should improve this to use topological sort or similar, so we resolve them in the most efficient order, and we don't have to do multiple iterations. Or maybe we can just loop in a defined order and then loop again backwards?

benhoyt:
Fixes #98

### DIFF
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -190,21 +190,30 @@ function set(a, k, v) {
 	}
 }
 
-func TestResolveTooManyIterations(t *testing.T) {
+func TestResolveLargeCallGraph(t *testing.T) {
+	const numCalls = 10000
+
 	var buf bytes.Buffer
 	var i int
-	for i = 0; i < 1200; i++ { // at least twice as big as maxResolveIterations in resolve.go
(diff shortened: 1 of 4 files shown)


vegarsti:
Cool!

PR #124, opened 2022-05-24
Make sure ctx being cancelled actually stops system() process

benhoyt:
This is a manifestation of https://github.com/golang/go/issues/21922 - use pipes to fix it. Tricky!


benhoyt:
Closing in favour of #270

PR #132, opened 2022-06-05
Interpret escapes in -v and var=value arguments

### LINKED ISSUE #129: -v must interpret escape sequences
Hi, @benhoyt.

`-v` must interpret escape sequences. I hate this spec of awk, but anyway, reality is what it is. ☹️

```console
$ goawk -v VAR='\t' 'BEGIN{printf VAR}' | hexdump -C
00000000  5c 74                                             |\t|

$ mawk -v VAR='\t' 'BEGIN{printf VAR}' | hexdump -C
00000000  09                                                |.|

$ gawk -v VAR='\t' 'BEGIN{printf VAR}' | hexdump -C
00000000  09                                                |.|
```

benhoyt:
This is odd, and doesn't seem to be explicitly stated in the POSIX spec, but it's how original-awk, mawk, and gawk all work. Portability trumps sanity. :-)

Fixes #129

### DIFF
--- a/goawk.go
+++ b/goawk.go
@@ -35,6 +35,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/pprof"
+	"strconv"
 	"strings"
 	"unicode/utf8"
 
@@ -269,11 +270,17 @@ func main() {
 		Output: stdout,
 	}
 	for _, v := range vars {
-		parts := strings.SplitN(v, "=", 2)
-		if len(parts) != 2 {
+		equals := strings.IndexByte(v, '=')
+		if equals < 0 {
 			errorExitf("-v flag must be in format name=value")
 		}
-		config.Vars = append(config.Vars, parts[0], parts[1])
+		name, value := v[:equals], v[equals+1:]
+		// Oddly, -v must interpret escapes (issue #129)
+		unquoted, err := strconv.Unquote(`"` + value + `"`)
+		if err == nil {
+			value = unquoted
+		}
+		config.Vars = append(config.Vars, name, value)
 	}
 
 	if cpuprofile != "" {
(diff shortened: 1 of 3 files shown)


ko1nksm:
Thanks!

> This is odd, and doesn't seem to be explicitly stated in the POSIX spec, 

I wish I had provided this link.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html

> assignment
An operand that begins with an <underscore> or alphabetic character from the portable character set (see the table in XBD [Portable Character Set](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap06.html#tag_06_01)), followed by a sequence of underscores, digits, and alphabetics from the portable character set, followed by the '=' character, shall specify a variable assignment rather than a pathname. The characters before the '=' represent the name of an awk variable; if that name is an awk reserved word (see [Grammar](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html#tag_20_06_13_16)) the behavior is undefined. The characters following the <equals-sign> shall be interpreted as if they appeared in the awk program preceded and followed by a double-quote ( '"') character, as a STRING token (see [Grammar](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html#tag_20_06_13_16)), except that if the last character is an unescaped <backslash>, it shall be interpreted as a literal <backslash> rather than as the first character of the sequence "\"". The variable shall be assigned the value of that STRING token and, if appropriate, shall be considered a numeric string (see [Expressions in awk](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html#tag_20_06_13_02)), the variable shall also be assigned its numeric value. Each such variable assignment shall occur just prior to the processing of the following file, if any. Thus, an assignment before the first file argument shall be executed after the BEGIN actions (if any), while an assignment after the last file argument shall occur before the END actions (if any). If there are no file arguments, assignments shall be executed before processing the standard input.

benhoyt:
Ah, I'd missed that, thanks! I read the section on `-v` but didn't realize that is was referring to that section. I'll update it to handle escapes as per the AWK lexer, and properly handle the backslash-at-the-end issue, rather than using Go's `strconv.Unquote`.

PR #140, opened 2022-07-14
Add support for -E argument (like Gawk's -E/--exec)

### LINKED ISSUE #139: Consider adding -Wexec option of mawk and gawk
`mawk` and `gawk` provide the `-Wexec` option (also written `-W exec`, and further provided by `gawk` as `-E` and `--exec`) that allows writing Awk scripts that look like standard command-line utilities by suitably processing `ARGV`. The problem with `#!/usr/bin/env -S awk -f` is that any hyphen-prefixed options provided on the command-line may be processed by the underlying Awk interpreter rather than the script (`gawk` passes along options it doesn't understand). The only way around this is to wrap the Awk call in a shell script that contains `exec env awk -f /path/to/script.awk -- "$@"`. This is what `prmindent` does in its [Makefile](https://git.sr.ht/~paulapatience/prmindent/tree/main/item/Makefile#L73). `-Wexec FILE` is effectively `-f FILE --`.

Why `-Wexec` rather than just `--exec`? From what I can gather from the `gawk` manual page, the POSIX-compliant way of adding implementation-specific options is by prefixing them with `-W`. Whether `goawk` should provide its other implementation-specific options under `-W` (and possibly only under `-W`, as `mawk` does) is something to be considered separately. In order to maximize compatibility with `mawk`, this option should be provided at least as `-Wexec` and `-W exec`.

The hashbang line of `prmindent` is `#!/usr/bin/env -S awk -Wexec` by default. This change would allow `goawk` to be a drop-in replacement for the system `awk` at least in this case.

benhoyt:
It's like -f but stops processing other arguments and disables
foo=bar arguments in the rest of the args. Like Gawk's -E/--exec and
mawk's "-W exec" option.

Also update test runner to run Gawk and GoAWK tests in separate
sub-tests, so one breaking doesn't stop the other from running.

Fixes #139 (thanks @paulapatience)

### DIFF
--- a/goawk.go
+++ b/goawk.go
@@ -57,6 +57,7 @@ Additional GoAWK arguments:
   -d                print parsed syntax tree to stderr (debug mode)
   -da               print virtual machine assembly instructions to stderr
   -dt               print variable type information to stderr
+  -E progfile       load program, treat as last option, disable var=value args
   -H                parse header row and enable @"field" in CSV input mode
   -h, --help        show this help message
   -i mode           parse input into fields using CSV format (ignore FS and RS)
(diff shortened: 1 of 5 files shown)


paulapatience:
Looks good to me.

Issue #144, opened 2022-08-19
Code coverage for GoAWK


xonixx:
In [this branch](https://github.com/benhoyt/goawk/compare/master...xonixx:goawk:coverage) I'm working on code coverage functionality for GoAWK, similar to the one built in the Golang itself (https://go.dev/blog/cover).

In fact, I was able to make the `go tool cover` to produce the miningful HTML coverage report for AWK. 

I wonder if you @benhoyt would be interested in adding this to the project. My solution is a bit hacky here and there, but the approach seems to work well. 

I'm still working on adding tests and documentation, and maybe some minor code adjustments, but I would be glad to start getting some feedback from you @benhoyt. 

![image](https://user-images.githubusercontent.com/11706893/185683039-e4d99308-264e-4410-83e4-6f04e20aba8c.png)

benhoyt:
Hi @xonixx, wow, this looks very interesting -- nice work. I'll take a closer look later this week and get back to you.

benhoyt:
I'll leave more comments later, but in the meantime I'm having trouble using `go tool cover` to view the report.

```
$ cat t.awk
BEGIN {
	if (1) {
		print "hello"
	}
	print "end"
}
$ go run . -d -covermode set -coverprofile cover.out -f t.awk
BEGIN {
    if (1) {
        __COVER[1] = 1
        print "hello"
    }
    __COVER[2] = 1
    print "end"
}
hello
end
$ cat cover.out
mode: set
t.awk:2.18,2.32 1 1
t.awk:2.18,2.32 1 1
t.awk:2.18,2.32 1 1
t.awk:2.18,2.32 1 1
t.awk:2.34,2.46 1 1
t.awk:2.18,2.32 1 1
t.awk:2.34,2.46 1 1
t.awk:2.18,2.32 1 1
t.awk:2.34,2.46 1 1
t.awk:2.18,2.32 1 1
t.awk:2.34,2.46 1 1
t.awk:3.3,3.16 1 1
t.awk:5.2,5.13 1 1
t.awk:3.3,3.16 1 1
t.awk:5.2,5.13 1 1
$ go tool cover -html=cover.out
cover: did not find package for t.awk in go list output
```

Do you know why? How did you run `go tool cover`?

xonixx:
[...]

PR #148, opened 2022-08-30
Decouple parse & resolve steps

xonixx:
This PR is prepared in context of #144.

It implements the part

> However, that would require separating out the resolver to a separate pass ... which is probably good to do anyway, but a bit of work.

Since the change is reasonably big I would add couple clarifications.

1. I tried to do only minimal work to get the task in question done, though I could not resist some (hopefully) minor refactorings and restructuring. 
2. It appears that resolving step needs some positions information, but not for all nodes. So at this point to keep the things simple and limited I wired the position only through very few of AST nodes. If need be we can equip all AST nodes with positions in generic way later but no need at this point.
3. I decided that we don't need to move ALL things that were in resolve.go to separate step. To me some things logically belong more to parsing than resolution. Example: `multiExprs` thing.
4. Since the parsing & resolving needed the same approach to errors with positions I generified this a bit by introducing the `PositionError` which is now panic-thrown internally by both parser & resolver. However I kept the external parser interface intact to return documented `ParseError`.
5. I checked the test suite - seems to work fine.


benhoyt:
Excellent, thanks for this. I'll review it -- hopefully over the next week or so. Note that I just merged another change, so there's a small conflict in `parser.go` which needs fixing.

xonixx:
> Note that I just merged another change, so there's a small conflict in parser.go which needs fixing.

Merged master.

xonixx:
@benhoyt Any update?

benhoyt:
Not yet, sorry. I glanced at it briefly, but am now traveling overseas. It
might be another week or two, but I will get to this.

xonixx:
Sure no problem. Have a pleasant trip!

internal/resolver/resolve.go
```diff
@@ -280,11 +354,11 @@ func (p *parser) resolveVars(prog *Program) {
...
 	prog.Scalars = make(map[string]int)
 	prog.Arrays = make(map[string]int)
-	for name, info := range p.varTypes[""] {
-		_, isFunc := p.functions[name]
+	for name, info := range r.varTypes[""] {
+		_, isFunc := r.functions[name]
 		if isFunc {
 			// Global var can't also be the name of a function
-			panic(p.errorf("global var %q can't also be a function", name))
+			panic(prog.EndPos.Errorf("global var %q can't also be a function", name))
```
benhoyt, line 361: This is only the program's "end pos" because we don't have access to a better position at this point. Ideally it'd be the position of the global var. For now, I think we should just make it position 1,1 (that's at last as good as the program's end pos) and remove `Program.EndPos`.

lexer/lexer.go
```diff
@@ -32,6 +33,32 @@ type Position struct {
 	Column int
 }
 
+func (pos Position) Add(lines int, cols int) Position {
```
benhoyt, line 36: I would rather we didn't add this `Add` method. It seems a bit weird to "add" to a position, and it's only used in one place, where I think it'd actually be preferable to make the error position the name itself, rather than just past the name.

internal/resolver/resolve.go
```diff
@@ -1,16 +1,107 @@
...
+	case *ast.VarExpr:
+		r.recordVarRef(n)
+
+	case *ast.ArrayExpr:
+		r.recordArrayRef(n)
+
+	case *ast.UserCallExpr:
+		name := n.Name
+		if r.locals[name] {
+			panic(n.Pos.Add(0, len(name)).Errorf("can't call local variable %q as function", name))
```
benhoyt, line 91: Let's make this error position just the position of the name, `n.Pos`. I think it's slightly better. It will mean updating a couple of error positions in tests.

lexer/lexer.go
```diff
@@ -32,6 +33,32 @@ type Position struct {
...
+	return Position{
+		Line:   pos.Line + lines,
+		Column: pos.Column + cols,
+	}
+}
+
+type PositionError struct {
```
benhoyt, line 43: Hmm, I can see why you've added this (for use in both parser and resolver), but I don't love how this is a new exported type in the lexer. It's only used internally. I think we should add this in package `ast` as `ast.PositionError` instead.

Additionally, I don't love the `Errorf` method being on the position type; I'd prefer it were just a method or function in the parser and/or resolver (it's a tiny function, so it's fine if it's duplicated in both if needed). Or you could add `ast.PosErrorf(pos, format, args...)` if you want to share it.

parser/parser.go
```diff
@@ -5,7 +5,7 @@
 package parser
 
 import (
-	"fmt"
+	"github.com/benhoyt/goawk/internal/resolver"
```
benhoyt, line 8: Let's move this down into a separate block (with a blank line in between stdlib imports and project imports).

parser/parser.go
```diff
@@ -45,35 +36,51 @@ type ParserConfig struct {
...
-		// errors internally, and they're caught here. This
-		// significantly simplifies the recursive descent calls as
-		// we don't have to check errors everywhere.
+		// The parser and resolver use panic with a *PositionError to signal parsing
+		// errors internally, and they're caught here. This significantly simplifies
+		// the recursive descent calls as we don't have to check errors everywhere.
 		if r := recover(); r != nil {
 			// Convert to ParseError or re-panic
-			err = r.(*ParseError)
+			err = &ParseError{*r.(*PositionError)}
```
benhoyt, line 60: Per comments in `lexer`, this will now be something like:

```
// Convert to *ast.PositionError or re-panic
posErr := r.(*ast.PositionError)
err = &ParseError{Position: posError.Position, Message: posErr.Message}
```

parser/parser.go
```diff
@@ -19,16 +19,7 @@ import (
...
-	Position Position
-	// Error message.
-	Message string
-}
-
-// Error returns a formatted version of the error, including the line
-// and column numbers.
-func (e *ParseError) Error() string {
-	return fmt.Sprintf("parse error at %d:%d: %s", e.Position.Line, e.Position.Column, e.Message)
+	PositionError
```
benhoyt, line 22: Because this will now be `ast.PositionError` (per comments in `lexer`), let's revert the change to these fields (mainly for the sake of API documentation), and just copy the fields over when we want to create the `ParseError`.

We'll need to add `ParseError.Error` back too.

parser/parser.go
```diff
@@ -45,35 +36,51 @@ type ParserConfig struct {
...
 // ParseProgram parses an entire AWK program, returning the *Program
 // abstract syntax tree or a *ParseError on error. "config" describes
 // the parser configuration (and is allowed to be nil).
 func ParseProgram(src []byte, config *ParserConfig) (prog *Program, err error) {
 	defer func() {
-		// The parser uses panic with a *ParseError to signal parsing
-		// errors internally, and they're caught here. This
-		// significantly simplifies the recursive descent calls as
-		// we don't have to check errors everywhere.
+		// The parser and resolver use panic with a *PositionError to signal parsing
```
benhoyt, line 55: `... panic with an *ast.PositionError ...`

parser/parser.go
```diff
@@ -1059,12 +1028,37 @@ func (p *parser) userCall(name string, pos Position) *ast.UserCallExpr {
...
 		i++
 	}
 	p.expect(RPAREN)
-	call := &ast.UserCallExpr{false, -1, name, args} // index is resolved later
-	p.recordUserCall(call, pos)
-	return call
+	return &ast.UserCallExpr{false, ast.WillBeResolvedLater, name, args, pos}
+}
+
+// Record a "multi expression" (comma-separated pseudo-expression
```
benhoyt, line 1038: Yeah, I agree it makes sense to keep the MultiExpr stuff in the parser -- thanks.

internal/resolver/resolve.go
```diff
@@ -180,98 +253,99 @@ func (p *parser) getScope(name string) (ast.VarScope, string) {
...
+	scope, funcName := r.getScope(name)
 	if scope == ast.ScopeSpecial {
-		panic(p.errorf("can't use scalar %q as array", name))
+		panic(expr.Pos.Errorf("can't use scalar %q as array", name))
 	}
-	expr := &ast.ArrayExpr{scope, 0, name}
-	p.arrayRefs = append(p.arrayRefs, arrayRef{funcName, expr, pos})
-	info := p.varTypes[funcName][name]
+	expr.Scope = scope
+	//expr := &ast.ArrayExpr{scope, 0, name}
```
benhoyt, line 276: Debugging comment left in?

parser/parser.go
```diff
@@ -83,19 +90,14 @@ type Program struct {
...
 	// be exported in package "parser", otherwise these could live in
 	// "internal/ast".)
-	Begin     []ast.Stmts
-	Actions   []ast.Action
-	End       []ast.Stmts
-	Functions []ast.Function
-	Scalars   map[string]int
-	Arrays    map[string]int
-	Compiled  *compiler.Program
+	ast.ResolvedProgram
```
benhoyt, line 93: At first I was a little hesitant to change this (as `Program` is exported), but I think I've convinced myself it's okay, especially given the above comment. Plus, the fields are still accessible (via the embedded `ResolvedProgram` struct) if anyone really is using them (they shouldn't be).

benhoyt (changes requested):
Thanks for your work here. I think it's really good work overall. I've left a bunch of comments, mostly minor and hopefully easy enough to address.

internal/ast/walk.go
```diff
@@ -0,0 +1,180 @@
...
+// A Visitor's Visit method is invoked for each node encountered by Walk.
+// If the result visitor w is not nil, Walk visits each of the children
+// of node with the visitor w, followed by a call of w.Visit(nil).
+type Visitor interface {
+	Visit(node Node) (w Visitor)
+}
+
+// Helper functions for common node lists. They may be empty.
+
+func WalkExprList(v Visitor, list []Expr) {
```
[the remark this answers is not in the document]
xonixx: This is true for `WalkExprList` but `WalkStmtList` is used in https://github.com/xonixx/goawk/blob/decouple_parse_resolve/internal/resolver/resolve.go#L77

parser/parser.go
xonixx: Ok so I assume we don't need to change anything here. Resolving then without changes. Please let me know if you think otherwise.

internal/resolver/resolve.go
```diff
@@ -1,16 +1,107 @@
...
+	r.resolveUserCalls(prog)
+	r.resolveVars(resolvedProg)
+
+	return resolvedProg
+}
+
+func (r *resolver) Visit(node ast.Node) ast.Visitor {
+	switch n := node.(type) {
+
+	case ast.Function:
```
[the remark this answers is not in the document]
xonixx: Done. Applied the same for `Action`.

lexer/lexer.go
xonixx: Yeah, I've added `ast.PosErrorf(pos, format, args...)`

parser/parser.go
xonixx: Sure. Strange, that gofmt doesn't take care of this.

xonixx:
Hey @benhoyt! Hopefully, I've resolved all items except for the one with `WalkExprList`/`WalkStmtList` - please see my comment there. What do you think?

internal/resolver/resolve.go
```diff
@@ -60,41 +151,28 @@ type varRef struct {
...
-
-// Signal the end of a function
-func (p *parser) stopFunction() {
-	p.funcName = ""
-}
-
-// Add function by name with given index
-func (p *parser) addFunction(name string, index int) {
-	p.functions[name] = index
+func (r *resolver) initResolve(config *Config) {
```
benhoyt, line 163: Nit: this could be named just `init` now that it's on the `resolver` type. In fact, what about making it a `newResolver(config *Config)` function instead?

parser/parser.go
```diff
@@ -45,35 +46,55 @@ type ParserConfig struct {
...
-		// errors internally, and they're caught here. This
-		// significantly simplifies the recursive descent calls as
-		// we don't have to check errors everywhere.
+		// The parser and resolver use panic with an *ast.PositionError to signal parsing
+		// errors internally, and they're caught here. This significantly simplifies
+		// the recursive descent calls as we don't have to check errors everywhere.
 		if r := recover(); r != nil {
 			// Convert to ParseError or re-panic
-			err = r.(*ParseError)
+			posError := *r.(*ast.PositionError)
```
benhoyt, line 70: Nit: `// Convert to PositionError or re-panic`

benhoyt (approved):
Great! Let's just fix these few comment nits and I'll squash and merge this!

xonixx:
Done.

benhoyt:
Thanks, merged!

PR #151, opened 2022-09-21
Draft version of FileReader

benhoyt:
This implements a simple `FileReader` type that tracks files added to it with `AddFile()` and their number of source lines, and then lets you look up which file and file-line number a particular overall line number belongs to with `FileLine()`.

It allows us to do away with `errorFileLine` and look up lines simply, without altering the parser or being invasive in that code.

Just a draft for comparing approaches. Needs comments and tests.

### DIFF
--- a/goawk.go
+++ b/goawk.go
@@ -30,7 +30,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"io/ioutil"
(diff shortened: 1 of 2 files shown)


benhoyt:
Closing in favour of #153

PR #153, opened 2022-09-22
Better filenames tracking

xonixx:
This is the other PR in context of #144.

This is re-implemenation of #150 based on work in #151. I think #151 can be just closed.

Basically this is almost equivalent to #151 plus:
1. added tests
2. added doc comments
3. applied minor optimization

### DIFF
--- a/internal/parseutil/filereader.go
+++ b/internal/parseutil/filereader.go
@@ -15,9 +15,8 @@ type FileReader struct {
 }
 
 type file struct {
-	path      string
-	startLine int // the number of first line of this file in joined source
-	lines     int
+	path  string
+	lines int
 }
 
 // AddFile adds a single source file.
@@ -33,21 +32,18 @@ func (fr *FileReader) AddFile(path string, source io.Reader) error {
 	}
 	content := fr.source.Bytes()[curLen:]
 	lines := bytes.Count(content, []byte("\n"))
-	startLine := 1
-	if len(fr.files) > 0 {
-		lastFile := fr.files[len(fr.files)-1]
-		startLine = lastFile.startLine + lastFile.lines
-	}
-	fr.files = append(fr.files, file{path, startLine, lines})
+	fr.files = append(fr.files, file{path, lines})
 	return nil
 }
 
 // FileLine resolves global line number in joined source to a local line number in a file (identified by path)
 func (fr *FileReader) FileLine(line int) (path string, fileLine int) {
+	startLine := 1
 	for _, f := range fr.files {
-		if line >= f.startLine && line < f.startLine+f.lines {
-			return f.path, line - f.startLine + 1
+		if line >= startLine && line < startLine+f.lines {
+			return f.path, line - startLine + 1
 		}
+		startLine += f.lines
 	}
 	return "", 0
 }
--- a/internal/parseutil/filereader.go
+++ b/internal/parseutil/filereader.go
@@ -49,7 +49,7 @@ func (fr *FileReader) FileLine(line int) (path string, fileLine int) {
 	return "", 0
 }
 
-// Source returns joined source of all input sources
+// Source returns the concatenated source code from all files added.
 func (fr *FileReader) Source() []byte {
 	return fr.source.Bytes()
 }
--- a/internal/parseutil/filereader.go
+++ b/internal/parseutil/filereader.go
@@ -7,7 +7,7 @@ import (
 )
 
 // FileReader serves two purposes:
-// 1. read input sources and join them for single source
+// 1. read input sources and join them into a single source (slice of bytes)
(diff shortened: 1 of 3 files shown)


xonixx:
@benhoyt Any thoughts?

benhoyt:
Yes, sorry, last week was a busy week. I'll try to get to this tonight or tomorrow.

internal/parseutil/filereader.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return err
+	}
+	if !bytes.HasSuffix(fr.source.Bytes(), []byte("\n")) {
+		fr.source.WriteByte('\n')
+	}
+	content := fr.source.Bytes()[curLen:]
+	lines := bytes.Count(content, []byte("\n"))
+	startLine := 1
+	if len(fr.files) > 0 {
```
benhoyt, line 35: I think we should undo the "optimization" in https://github.com/benhoyt/goawk/pull/153/commits/00df7cd1443e48aac033b9f8cf9bea07c16f1c33 -- i.e., remove the `file.startLine` field. It complicates this code slightly (a few more lines and the `if` statement) and only saves a single integer addition in `FileLine`.

internal/parseutil/filereader.go
```diff
@@ -0,0 +1,56 @@
...
+		lastFile := fr.files[len(fr.files)-1]
+		startLine = lastFile.startLine + lastFile.lines
+	}
+	fr.files = append(fr.files, file{path, startLine, lines})
+	return nil
+}
+
+// FileLine resolves global line number in joined source to a local line number in a file (identified by path)
```
benhoyt, line 43: Let's wrap the comment lines to <80 chars, and update the wording a little here:

```go
// FileLine resolves an overall line number from the concatenated source code
// to the local line number in that source file (identified by path).
```

internal/parseutil/filereader.go
```diff
@@ -0,0 +1,56 @@
...
+func (fr *FileReader) FileLine(line int) (path string, fileLine int) {
+	for _, f := range fr.files {
+		if line >= f.startLine && line < f.startLine+f.lines {
+			return f.path, line - f.startLine + 1
+		}
+	}
+	return "", 0
+}
+
+// Source returns joined source of all input sources
```
benhoyt, line 53: Let's tweak the wording a little:

```go
// Source returns the concatenated source code from all files added.
```

internal/parseutil/filereader_test.go
```diff
@@ -0,0 +1,48 @@
+package parseutil
```
benhoyt, line 1: This is a straight-forward "no dependencies" package, so let's put this in package `parseutil_test` so we're not testing internals, and only test the exported interface.

internal/parseutil/filereader_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestLineResolution(t *testing.T) {
```
benhoyt, line 8: Let's call this `TestFileReader` if we're testing all of `FileReader`, or break it up into smaller functions with more specific names if that's simpler.

internal/parseutil/filereader_test.go
```diff
@@ -0,0 +1,48 @@
...
+	fr := &FileReader{}
+
+	file1 := "file1"
+	file2 := "file2"
```
benhoyt, line 12: I think these constants obscure the test rather than clarify. Let's just inline them.

internal/parseutil/filereader_test.go
```diff
@@ -0,0 +1,48 @@
...
+print f(1)
+}`)
+	addFile(file2, `function f(x) {
+print x
+}`)
+	if len(fr.files) != 2 {
+		t.Errorf("must be 2 files")
+	}
+
+	{
```
benhoyt, line 30: Maybe we can make these table-driven subtests with `t.Run`?

I think we want a few more tests, specifically no files, a zero length file, and files with and without a trailing `\n` on the last line.

Perhaps a table-driven test struct something like this:

```go
type test struct {
    // input:
    files []struct{ name, source string }
    line int
    // expected:
    path string
    fileLine int
}
```

And make sure we test `Source()` for completeness.

benhoyt (changes requested):
Thanks for this. Looks good overall! Requesting a few minor changes and some additional test cases.

internal/parseutil/filereader.go
xonixx: Agree. Reverted.

xonixx:
Done. Please check if it's ok now.

benhoyt:
Merged, thanks!

PR #158, opened 2022-10-22
Coverage fix

xonixx:
One more PR for #144 that addresses couple of minor things, missed in #154 

In this PR:

1. Remove unintuitive behavior (outputting annotated code when only `-covermode` flag set) - just use `-d` flag
2. Update cover.png screenshot: now we show if/for/while conditions in green (covered)
3. Fix for `./goawk -covermode=set 'BEGIN { print 123 }'` to not hang waiting for stdin. This started to happen because processing stops if `prog.Actions == nil` and we by mistake were changing it to empty slice during coverage annotation.


benhoyt (approved):
Thanks! I noticed the "waiting for stdin" issue while testing yesterday as well. I think your fix is fine, but also added a commit which I think "fixes the issue at its source". I was using `program.Actions` and `program.End` directly, instead of `program.Compiled.Actions` and `program.Compiled.End`. (I also made the tests `len(slice)==0` rather than `slice==nil` while in there.

Issue #205, opened 2023-08-27
system() always returns -1 if process exits due to signal


juster:
Another incompatible feature of awks is the return value of `system()` when death by signal is involved. The [POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html#tag_20_06_13_14) stipulates that the results of awk's `system()` should be the value of C's `system(3)` (which is the value of `waitpid(2)`). But due to possible human error the value of awk's `system()` on signal exit was originally a fraction! This is C's `waitpid(2)` result divided by 256.

The source here is goawk's `system()` [docs](https://www.gnu.org/software/gawk/manual/gawk.html#index-system_0028_0029-function):

> On POSIX systems, a command’s exit status is a 16-bit number. The exit value passed to the C exit() function is held in the high-order eight bits. The low-order bits indicate if the process was killed by a signal (bit 7) and if so, the guilty signal number (bits 0–6).
> 
> Traditionally, awk’s system() function has simply returned the exit status value divided by 256. In the normal case this gives the exit status but in the case of death-by-signal it yields a fractional floating-point value.[52](https://www.gnu.org/software/gawk/manual/gawk.html#FOOT52) POSIX states that awk’s system() should return the full 16-bit value.
>
> gawk steers a middle ground. The return values are summarized in [Table 9.5](https://www.gnu.org/software/gawk/manual/gawk.html#table_002dsystem_002dreturn_002dvalues).

### Example test command

```
goawk 'BEGIN { print system("kill -1 $$") }'
```

### awk variant output

| Variant | Output | Matches close(pipe)? | Comment |
| ------- | ------- | --------------------- | ---------- |
| goawk | -1 | N | Any signal always causes -1 |
| lucent | 0.00390625 | N | 1 / 256 |
| nawk | " | " | " |
| gawk | 257 | Y | 256 + 1 |
| mawk | " | " | " |
| POSIX standard | 1 | N | Result of `waitpid()` |

### Summary

There is no right answer. You *could* be the first awk to implement the POSIX standard ;) but I would suggest following gawk's example.

Another nice thing gawk did was to make the result the same as `close()` on a command pipe. Original awk behavior isn't *terrible* here but it forgets to match `system()` with the `close()` on a pipe. This may be due to the fact that using the `close()` result on pipe was not implemented in original awk.

I stumbled across this when matching gawk's behavior for #203. I will make both `close(pipe)` and `system()` match gawk's result in my forked branch, for now.

Issue #204, opened 2023-08-25
close() on input pipe does not cleanup child process


juster:
When `close()` is called on an input pipe, the `close()` builtin function does not `.Wait` for the `exec.Cmd` at the other end of the "pipe" and so the resources associated with the child process are not released. This leaves "zombie processes" which continue to be tracked in the parent, and maybe store their return value. `close()` will cause them to be forgotten by the interpreter so new processes will be spawned for the same command.

### Example

You can easily reach reach the limit on the number of open processes, at which point the fork/exec call no longer works. Here, I set the limit to 512 (pretty low) and give it a go.

```
$ ulimit -u 512
$ goawk 'BEGIN { for(i=1;i<=512;i++){ "date"|getline; close("date") }}'
fork/exec /bin/sh: resource temporarily unavailable
fork/exec /bin/sh: resource temporarily unavailable
fork/exec /bin/sh: resource temporarily unavailable
...

juster:
I thought it would be nice to have interfaces for input/output streams and specific structs for file/pipe streams. I still have to make another pass to clean things up and also write tests. But the commit above fixed my problem here and the minor issue in #203.

benhoyt:
This is interesting work, thanks @juster. Cleaning up the structure of handling file/pipe streams is something I've been hoping to do for a while -- I half suspected this was an issue but have never looked into it. So thanks. I'll look at this more closely (including your commit) in the next couple of weeks.

benhoyt:
Out of interest, @juster: what are you using GoAWK for, and how did you stumble across this issue? I'm always interested to learn more about how it's being used.

juster:
I use goawk all the time for analyzing server logs. They can be in the GBs. A coworker suggested goawk because it parses CSV. I used to use a lot of Perl or python but I decided to experiment if awk was capable of more or less the same thing. Goawk is also faster than nawk and often even grep!

One weakness with awk is with time stamps. I found this bug because I was trying to parse time stamps with the date command! So I was opening quite a few commands.

Regardless of this bug it was too slow. But then I discovered I could provide custom awk functions written in Go and embed a script in a binary. I love it!

Issue #203, opened 2023-08-25
close() on pipe command does not return exit status


juster:
An undocumented feature of other awks is to have the `close()` builtin function return the exit status of the process which was opened by the pipe. The `close()` builtin for _goawk_ always returns _-1_ for output pipes and always returns _0_ for input pipes.

### Example

```
$ goawk 'BEGIN { print|"exit 1"; print close("exit 1") }'
-1
$ goawk 'BEGIN { "exit 1"|getline; print close("exit 1") }'
0
```

### Portability?

Different awks also have various implementations. The same example above was run for ...

1. the "one-true" Lucent copyright awk
2. GNU gawk
3. plan9 from userspace awk
4. Mac OS awk

| Variant | Input/Output Pipe | Behavior |
| ------- | ------------- | --------- |
| one-true / mac os | both | Value is 256 (1 << 8) |
| gawk | both | Value is 1 |
| plan9 | both | Use of the value of close() is illegal |

edit: Removed duplicate rows. I also remembered how the `wait()` result shifts the exit code.

benhoyt:
Interesting! Linking the [GAWK docs on this](https://www.gnu.org/software/gawk/manual/gawk.html#Close-Return-Value) for reference. For comparison, POSIX just says "If the close was successful, the function shall return zero; otherwise, it shall return non-zero."

juster:
Ah nice! Those linked docs explain plan9 awk's behavior:

> In many older versions of Unix awk, the close() function is actually a statement. (d.c.) It is a syntax error to try and use the return value from close(): 

It complains about an illegal statement.

But yeah this is definitely an undocumented easter egg!

PR #206, opened 2023-08-30
Abstract redirects and pipes into inputStream and outputStream interface

juster:
This creates the `inputStream` and `outputStream` interfaces which are satisfied by structs for all combinations of file/pipe and input/output:

1. Input File - `struct inFileStream`
2. Input Command - `struct inCmdStream`
3. Output File - `struct outFileStream`
4. Output Command - `struct outCmdStream`

The streams satisfy the `io.ReadCloser` and `io.WriteCloser` interfaces while the `outputStream` also has a `Flush() error` method. Both types of streams also have an `ExitCode() int` method which makes it trivial to determine the result of a `close()` on those streams.

This PR homogenizes behavior and fixes a number of minor issues related to command pipes. Gawk behavior is mimicked for the following:

1. The exit status of the child process is returned from `close()` when closing a pipe. Exit due to signal is `256 + signal` and if both a signal and coredump occurred it is `512 + signal`. (see #203)
2. The same rules apply to the result of the `system()` builtin function. (see #205)

A separate bug was fixed so that the child process information is cleaned up when `close()` is called on an input pipe (i.e. via `getline`). (see #204)

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -587,14 +587,14 @@ BEGIN {
 		"", "error\n0\n", "", ""},
 	{`BEGIN { print system("exit 42") }  # !fuzz !posix`, "", "42\n", "", ""},
 	{`BEGIN { system("cat") }`, "foo\nbar", "foo\nbar", "", ""},
-	{`BEGIN { print system("exec kill -9 $$") } # !awk !posix !windows-gawk`, "", "265\n", "", ""},
+	{`BEGIN { print system("exec /bin/kill -9 $$") } # !awk !posix !windows`, "", "265\n", "", ""},
 
 	// Test bytes/unicode handling (GoAWK currently has char==byte, unlike Gawk).
-	{`BEGIN { print match("food", "foo"), RSTART, RLENGTH }  !gawk`, "", "1 1 3\n", "", ""},
-	{`BEGIN { print match("x food y", "fo"), RSTART, RLENGTH }  !gawk`, "", "3 3 2\n", "", ""},
-	{`BEGIN { print match("x food y", "fox"), RSTART, RLENGTH }  !gawk`, "", "0 0 -1\n", "", ""},
-	{`BEGIN { print match("x food y", /[fod]+/), RSTART, RLENGTH }  !gawk`, "", "3 3 4\n", "", ""},
-	{`BEGIN { print match("絵 fööd y", /[föd]+/), RSTART, RLENGTH }  !gawk`, "", "5 5 6\n", "", ""},
+	{`BEGIN { print match("food", "foo"), RSTART, RLENGTH } # !gawk`, "", "1 1 3\n", "", ""},
+	{`BEGIN { print match("x food y", "fo"), RSTART, RLENGTH } # !gawk`, "", "3 3 2\n", "", ""},
+	{`BEGIN { print match("x food y", "fox"), RSTART, RLENGTH } # !gawk`, "", "0 0 -1\n", "", ""},
+	{`BEGIN { print match("x food y", /[fod]+/), RSTART, RLENGTH } # !gawk`, "", "3 3 4\n", "", ""},
+	{`BEGIN { print match("絵 fööd y", /[föd]+/), RSTART, RLENGTH } # !gawk`, "", "5 5 6\n", "", ""},
 	{`{ print length, length(), length("buzz"), length("") }  # !gawk`, "foo bar", "7 7 4 0\n", "", ""},
 	{`BEGIN { print length("a"), length("絵") }  # !gawk`, "", "1 3\n", "", ""},
 	{`BEGIN { print index("foo", "f"), index("foo0", 0), index("foo", "o"), index("foo", "x") }  # !gawk`, "", "1 4 2 0\n", "", ""},
@@ -844,9 +844,9 @@ BEGIN { x[1]=3; f5(x); print x[1] }
 	// The value of close() on a pipe emulates gawk behavior. Results are identical for both
 	// input and output. Windows does not do POSIX signals.
 	{`BEGIN { cmd="exit 9"; print "" |cmd; print close(cmd) } # !awk !posix`, "", "9\n", "", ""},
-	{`BEGIN { cmd="exec kill -9 $$"; print "" |cmd; print close(cmd) } # !awk !posix !windows-gawk`, "", "265\n", "", ""},
+	{`BEGIN { cmd="exec /bin/kill -9 $$"; print "" |cmd; print close(cmd) } # !awk !posix !windows`, "", "265\n", "", ""},
 	{`BEGIN { cmd="exit 9"; cmd |getline; print close(cmd) } # !awk !posix`, "", "9\n", "", ""},
-	{`BEGIN { cmd="exec kill -9 $$"; cmd |getline; print close(cmd) } # !awk !posix !windows-gawk`, "", "265\n", "", ""},
+	{`BEGIN { cmd="exec /bin/kill -9 $$"; cmd |getline; print close(cmd) } # !awk !posix !windows`, "", "265\n", "", ""},
 
 	// Redirecting to or from a filename of "-" means write to stdout or read from stdin
 	{`BEGIN { print getline x < "-"; print x }`, "a\nb\n", "1\na\n", "", ""},
@@ -991,6 +991,28 @@ BEGIN { x[1]=3; f5(x); print x[1] }
 	{`{ print $1<$2 }`, "1_0 2", "1\n", "", ""},
 }
 
+// Flags parses the "skip flags" in it.src's first comment and returns them as
+// a map. An enable flag ("# +flag") appears as {"flag": true} in the map, and
+// a disable flag ("# !flag") appears as {"flag": false} in the map.
+func (it interpTest) Flags() map[string]bool {
+	flags := make(map[string]bool)
+	i := strings.Index(it.src, "#")
+	if i < 0 {
+		return flags
+	}
+	for _, field := range strings.Fields(it.src[i+1:]) {
+		switch {
+		case len(field) <= 1:
+			continue
+		case field[0] == '!':
+			flags[field[1:]] = false
+		case field[0] == '+':
+			flags[field[1:]] = true
+		}
+	}
+	return flags
+}
+
 func TestInterp(t *testing.T) {
 	// Ensure very long lines work (> 64KB)
 	longLine := strings.Repeat("x", 70000)
@@ -1003,20 +1025,30 @@ func TestInterp(t *testing.T) {
 		if len(testName) > 70 {
 			testName = testName[:70]
 		}
+		flags := test.Flags()
+		var skipWindows bool
+		if f, ok := flags["windows"]; !f && ok && runtime.GOOS == "windows" {
+			skipWindows = true
+		}
 
 		// Run it through external awk program first
 		if awkExe != "" {
 			runAWK := func(t *testing.T, posix bool) {
-				if strings.Contains(test.src, "!"+awkExe) {
+				if skipWindows {
+					t.Skip("skipping windows")
+				}
+
+				var f, ok bool
+				if f, ok = flags[awkExe]; !f && ok {
 					t.Skipf("skipping under %s", awkExe)
 				}
-				if strings.Contains(test.src, "!"+runtime.GOOS+"-"+awkExe) {
+				if f, ok = flags[runtime.GOOS+"-"+awkExe]; !f && ok {
 					t.Skipf("skipping on %s under %s", runtime.GOOS, awkExe)
 				}
-				if posix && strings.Contains(test.src, "!posix") {
+				if f, ok = flags["posix"]; posix && !f && ok {
 					t.Skipf("skipping in --posix mode")
 				}
-				if !posix && strings.Contains(test.src, "+posix") {
+				if !posix && f {
 					t.Skip("skipping in non-posix mode")
 				}
 
@@ -1063,6 +1095,9 @@ func TestInterp(t *testing.T) {
 
 		// Then test it in GoAWK
 		t.Run(testName, func(t *testing.T) {
+			if skipWindows {
+				t.Skip("skipping windows")
+			}
 			testGoAWK(t, test.src, test.in, test.out, test.err, nil, nil)
 		})
 	}
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -587,6 +587,7 @@ BEGIN {
 		"", "error\n0\n", "", ""},
 	{`BEGIN { print system("exit 42") }  # !fuzz !posix`, "", "42\n", "", ""},
 	{`BEGIN { system("cat") }`, "foo\nbar", "foo\nbar", "", ""},
+	{`BEGIN { print system("exec kill -9 $$") } # !awk !posix !windows-gawk`, "", "265\n", "", ""},
 
 	// Test bytes/unicode handling (GoAWK currently has char==byte, unlike Gawk).
 	{`BEGIN { print match("food", "foo"), RSTART, RLENGTH }  !gawk`, "", "1 1 3\n", "", ""},
@@ -839,6 +840,13 @@ BEGIN { x[1]=3; f5(x); print x[1] }
 	{`BEGIN { print |"out"; getline <"out" }  # !awk !gawk`, "", "", "can't read from writer stream", ""},
 	{`BEGIN { print >"out"; close("out"); getline <"out"; print >"out" }  # !awk !gawk`, "", "", "can't write to reader stream", ""},
 	{`BEGIN { print >"out"; close("out"); getline <"out"; print |"out" }  # !awk !gawk`, "", "", "can't write to reader stream", ""},
+
+	// The value of close() on a pipe emulates gawk behavior. Results are identical for both
+	// input and output. Windows does not do POSIX signals.
+	{`BEGIN { cmd="exit 9"; print "" |cmd; print close(cmd) } # !awk !posix`, "", "9\n", "", ""},
+	{`BEGIN { cmd="exec kill -9 $$"; print "" |cmd; print close(cmd) } # !awk !posix !windows-gawk`, "", "265\n", "", ""},
+	{`BEGIN { cmd="exit 9"; cmd |getline; print close(cmd) } # !awk !posix`, "", "9\n", "", ""},
+	{`BEGIN { cmd="exec kill -9 $$"; cmd |getline; print close(cmd) } # !awk !posix !windows-gawk`, "", "265\n", "", ""},
 
 	// Redirecting to or from a filename of "-" means write to stdout or read from stdin
 	{`BEGIN { print getline x < "-"; print x }`, "a\nb\n", "1\na\n", "", ""},
--- a/interp/iostream.go
+++ b/interp/iostream.go
@@ -1,0 +1,224 @@
+package interp
+
+// I/O streams are interfaces which allow file redirects and command pipelines to be treated
+// equivalently.
+
+import (
+	"bufio"
+	"errors"
+	"io"
+	"os/exec"
+	"syscall"
+)
+
+const (
+	notClosedExitCode = -127
+)
+
+var (
+	doubleCloseError = errors.New("close: stream already closed")
+)
+
+// firstError returns the first non-nil error or nil if all errors are nil.
+func firstError(errs ...error) error {
+	for _, err := range errs {
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Close the cmd and convert the error result into the result returned from goawk builtin functions.
+// A nil error is returned if that error describes a non-zero exit status or an unhandled signal.
+// Any other type of error returns -1 and err.
+//
+// The result mimicks gawk for expected child process errors:
+// 1. Returns the exit status of the child process and nil error on normal process exit.
+// 2. Returns 256 + signal on unhandled signal exit.
+// 3. Returns 512 + signal on unhandled signal exit which caused a core dump.
+func waitExitCode(cmd *exec.Cmd) (int, error) {
+	err := cmd.Wait()
+	if err == nil {
+		return 0, nil
+	}
+	ee, ok := err.(*exec.ExitError)
+	if !ok {
+		// Wait() returned an io error.
+		return -1, err
+	}
+	status, ok := ee.ProcessState.Sys().(syscall.WaitStatus)
+	if !ok {
+		// Maybe not all platforms support WaitStatus?
+		return -1, err
+	}
+	switch {
+	case status.CoreDump():
+		return 512 + int(status.Signal()), nil
+	case status.Signaled():
+		return 256 + int(status.Signal()), nil
+	case status.Exited():
+		return status.ExitStatus(), nil
+	default:
+		return -1, err
+	}
+}
+
+type inputStream interface {
+	io.ReadCloser
+	ExitCode() int
+}
+
+type outputStream interface {
+	io.WriteCloser
+	Flush() error
+	ExitCode() int
+}
+
+type outFileStream struct {
+	*bufio.Writer
+	closer   io.Closer
+	exitCode int
+	closed   bool
+}
+
+func newOutFileStream(wc io.WriteCloser, size int) outputStream {
+	b := bufio.NewWriterSize(wc, size)
+	return &outFileStream{b, wc, notClosedExitCode, false}
+}
+
+func (s *outFileStream) Close() error {
+	if s.closed {
+		return doubleCloseError
+	}
+	s.closed = true
+	flushErr := s.Writer.Flush()
+	closeErr := s.closer.Close()
+	if err := firstError(flushErr, closeErr); err != nil {
+		s.exitCode = -1
+		return err
+	}
+	s.exitCode = 0
+	return nil
+}
+
+func (s *outFileStream) ExitCode() int {
+	return s.exitCode
+}
+
+type outCmdStream struct {
+	*bufio.Writer
+	closer   io.Closer
+	cmd      *exec.Cmd
+	exitCode int
+	closed   bool
+}
+
+func newOutCmdStream(cmd *exec.Cmd) (outputStream, error) {
+	w, err := cmd.StdinPipe()
+	if err != nil {
+		return nil, newError("error connecting to stdin pipe: %v", err)
+	}
+	err = cmd.Start()
+	if err != nil {
+		w.Close()
+		return nil, err
+	}
+	out := &outCmdStream{bufio.NewWriterSize(w, outputBufSize), w, cmd, notClosedExitCode, false}
+	return out, nil
+}
+
+func (s *outCmdStream) Close() error {
+	if s.closed {
+		return doubleCloseError
+	}
+	s.closed = true
+	flushErr := s.Writer.Flush()
+	closeErr := s.closer.Close()
+	var waitErr error
+	s.exitCode, waitErr = waitExitCode(s.cmd)
+	return firstError(waitErr, flushErr, closeErr)
+}
+
+func (s *outCmdStream) ExitCode() int {
+	return s.exitCode
+}
+
+// An outNullStream allows writes to not do anything while fulfilling the outputStream interface.
+type outNullStream struct {
+	io.Writer
+	closed bool
+}
+
+func newOutNullStream() outputStream { return &outNullStream{io.Discard, false} }
+func (s outNullStream) Flush() error { return nil }
+func (s *outNullStream) Close() error {
+	if s.closed {
+		return doubleCloseError
+	}
+	s.closed = true
+	return nil
+}
+func (s outNullStream) ExitCode() int { return -1 }
+
+type inFileStream struct {
+	io.ReadCloser
+	exitCode int
+	closed   bool
+}
+
+func newInFileStream(rc io.ReadCloser) inputStream {
+	return &inFileStream{rc, notClosedExitCode, false}
+}
+
+func (s *inFileStream) Close() error {
+	if s.closed {
+		return doubleCloseError
+	}
+	s.closed = true
+	if err := s.ReadCloser.Close(); err != nil {
+		s.exitCode = -1
+		return err
+	}
+	s.exitCode = 0
+	return nil
+}
+
+func (s *inFileStream) ExitCode() int {
+	return s.exitCode
+}
+
+type inCmdStream struct {
+	io.ReadCloser
+	cmd      *exec.Cmd
+	exitCode int
+	closed   bool
+}
+
+func newInCmdStream(cmd *exec.Cmd) (inputStream, error) {
+	r, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, newError("error connecting to stdout pipe: %v", err)
+	}
+	err = cmd.Start()
+	if err != nil {
+		r.Close()
+		return nil, err
+	}
+	return &inCmdStream{r, cmd, notClosedExitCode, false}, nil
+}
+
+func (s *inCmdStream) Close() error {
+	if s.closed {
+		return doubleCloseError
+	}
+	s.closed = true
+	closeErr := s.ReadCloser.Close()
+	var waitErr error
+	s.exitCode, waitErr = waitExitCode(s.cmd)
+	return firstError(waitErr, closeErr)
+}
+
+func (s *inCmdStream) ExitCode() int {
+	return s.exitCode
+}
--- a/interp/iostream_test.go
+++ b/interp/iostream_test.go
@@ -1,0 +1,73 @@
+package interp
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+)
+
+const (
+	outputStreamBufferSize = 1024
+)
+
+func TestStreamDoubleClose(t *testing.T) {
+	dir := t.TempDir()
+	t.Run("InFile", func(t *testing.T) {
+		f, err := os.Create(filepath.Join(dir, "infile"))
+		if err != nil {
+			t.Fatal(err)
+		}
+		in := newInFileStream(f)
+		checkDoubleClose(t, in)
+	})
+	t.Run("OutFile", func(t *testing.T) {
+		f, err := os.Create(filepath.Join(dir, "outfile"))
+		if err != nil {
+			t.Fatal(err)
+		}
+		out := newOutFileStream(f, outputStreamBufferSize)
+		checkDoubleClose(t, out)
+	})
+	t.Run("InCmd", func(t *testing.T) {
+		cmd := execDefaultShell("echo close me")
+		in, err := newInCmdStream(cmd)
+		if err != nil {
+			t.Fatal(err)
+		}
+		checkDoubleClose(t, in)
+	})
+	t.Run("OutCmd", func(t *testing.T) {
+		cmd := execDefaultShell("echo close me")
+		out, err := newOutCmdStream(cmd)
+		if err != nil {
+			t.Fatal(err)
+		}
+		checkDoubleClose(t, out)
+	})
+}
+
+func execDefaultShell(scriptlet string) *exec.Cmd {
+	cmdline := append(defaultShellCommand, scriptlet)
+	return exec.Command(cmdline[0], cmdline[1:]...)
+}
+
+type streamCloser interface {
+	ExitCode() int
+	Close() error
+}
+
+func checkDoubleClose(t *testing.T, sc streamCloser) {
+	t.Helper()
+	if err := sc.Close(); err != nil {
+		t.Fatal(err)
+	}
+	exitCode := sc.ExitCode()
+	if err := sc.Close(); !errors.Is(err, doubleCloseError) {
+		t.Error("expected stream.Close() to return error on double close")
+	}
+	if sc.ExitCode() != exitCode {
+		t.Error("expected stream.ExitCode() to stay the same")
+	}
+}
--- a/interp/vm.go
+++ b/interp/vm.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"math"
 	"os"
-	"os/exec"
 	"strings"
 	"time"
 
@@ -1077,20 +1076,21 @@ func (p *interp) callBuiltin(builtinOp compiler.BuiltinOp) error {
 		cmd.Stdout = p.output
 		cmd.Stderr = p.errorOutput
 		_ = p.flushAll() // ensure synchronization
-		err := cmd.Run()
-		ret := 0.0
+		err := cmd.Start()
+		if err != nil {
+			// Could not start the shell so skip waiting on it.
+			p.printErrorf("%v\n", err)
+			p.replaceTop(num(-1.0))
+			return nil
+		}
+		exitCode, err := waitExitCode(cmd)
 		if err != nil {
 			if p.checkCtx && p.ctx.Err() != nil {
 				return p.ctx.Err()
 			}
-			if exitErr, ok := err.(*exec.ExitError); ok {
-				ret = float64(exitErr.ProcessState.ExitCode())
-			} else {
-				p.printErrorf("%v\n", err)
-				ret = -1
-			}
+			p.printErrorf("%v\n", err)
 		}
-		p.replaceTop(num(ret))
+		p.replaceTop(num(float64(exitCode)))
 
 	case compiler.BuiltinTolower:
 		p.replaceTop(str(strings.ToLower(p.toString(p.peekTop()))))
--- a/interp/vm.go
+++ b/interp/vm.go
@@ -911,33 +911,25 @@ func (p *interp) callBuiltin(builtinOp compiler.BuiltinOp) error {
 		p.replaceTop(num(math.Atan2(y.num(), x.num())))
 
 	case compiler.BuiltinClose:
+		var err error
+		code := -1
 		name := p.toString(p.peekTop())
-		var c io.Closer = p.inputStreams[name]
-		if c != nil {
+		if stream := p.inputStreams[name]; stream != nil {
 			// Close input stream
 			delete(p.inputStreams, name)
-			err := c.Close()
-			if err != nil {
-				p.replaceTop(num(-1))
-			} else {
-				p.replaceTop(num(0))
-			}
-		} else {
-			c = p.outputStreams[name]
-			if c != nil {
-				// Close output stream
-				delete(p.outputStreams, name)
-				err := c.Close()
-				if err != nil {
-					p.replaceTop(num(-1))
-				} else {
-					p.replaceTop(num(0))
-				}
-			} else {
-				// Nothing to close
-				p.replaceTop(num(-1))
-			}
+			delete(p.scanners, name)
+			err = stream.Close()
+			code = stream.ExitCode()
+		} else if stream := p.outputStreams[name]; stream != nil {
+			// Close output stream
+			delete(p.outputStreams, name)
+			err = stream.Close()
+			code = stream.ExitCode()
 		}
+		if err != nil {
+			p.printErrorf("error closing %q: %v\n", name, err)
+		}
+		p.replaceTop(num(float64(code)))
 
 	case compiler.BuiltinCos:
 		p.replaceTop(num(math.Cos(p.peekTop().num())))
--- a/goawk_test.go
+++ b/goawk_test.go
@@ -234,9 +234,6 @@ func TestGAWK(t *testing.T) {
 
 		"gsubtst7":     true, // something wrong with gsub or field split/join
 		"splitwht":     true, // other awks handle split(s, a, " ") differently from split(s, a, / /)
-		"status-close": true, // hmmm, not sure what's up here
-		"sigpipe1":     true, // probable race condition: sometimes fails, sometimes passes
-
 		"parse1": true, // incorrect parsing of $$a++++ (see TODOs in interp_test.go too)
 
 		"rscompat": true, // GoAWK allows multi-char RS by default
@@ -251,6 +248,7 @@ func TestGAWK(t *testing.T) {
 		"eofsplit": true, // reads from /etc/passwd
 		"getline5": true, // removes a file while it's open
 		"iobug1":   true, // reads from /dev/null
+		"sigpipe1": true, // requires POSIX signals
 	}
 
 	sortLines := map[string]bool{
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -20,7 +20,6 @@ import (
 	"math"
 	"math/rand"
 	"os"
-	"os/exec"
 	"regexp"
 	"runtime"
 	"strconv"
@@ -80,9 +79,8 @@ type interp struct {
 	hadFiles      bool
 	input         io.Reader
 	inputBuffer   []byte
-	inputStreams  map[string]io.ReadCloser
-	outputStreams map[string]io.WriteCloser
-	commands      map[string]*exec.Cmd
+	inputStreams  map[string]inputStream
+	outputStreams map[string]outputStream
 	noExec        bool
 	noFileWrites  bool
 	noFileReads   bool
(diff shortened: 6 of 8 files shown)


benhoyt:
Thanks for this -- this is excellent work! I hope to take a closer look at the code changes this weekend, but at first glance is looks like a good approach.

FYI, I've just merged https://github.com/benhoyt/goawk/pull/207 to update to Go 1.16 so we can use `io.Discard` etc now. So if you merge the latest master into this branch, that should fix the test failures.

juster:
That's great thanks! Do you mind if I rebase instead of merge? I rebase a lot it makes the commit history cleaner but spams up the issues & PRs... 😞 

I noticed that the Linux test had a failure. It seemed that the signal may have been handled and the sh process itself did not exit due to signal. I'll have to dig in more there and research which sh implementation is used and how it handles signals, etc. Ofc that is just a guess for now.

benhoyt:
@juster Yep, rebasing is fine. My development style is usually "push any old commit to the PR branch, then squash and merge when merging the PR", but if you keep a clean commit history in the PR, I'm happy to apply directly to master instead.

juster:
I found the problem with the `ExitSigPipe` test on Linux runners:

```
--- FAIL: TestStreamInCmd (0.00s)
    --- FAIL: TestStreamInCmd/ExitSigPipe (0.52s)
        iostream_test.go:168: expected Close() result: 269, got: 141
```

This test is running the `yes` utility because it prints non-stop. When I close the stream it will cause a `SIGPIPE` (13) signal to be sent to `yes` because it tries to print to a closed pipe.  The difference here is that `sh` is linked to `bash` and `bash` has a special exit code if one of its commands dies due to signal:

> When a command terminates on a fatal signal whose number is N, Bash uses the value 128+N as the exit status.

[Bash Reference Manual](https://www.gnu.org/software/bash/manual/bash.html#Exit-Status)

On my old Macbook `sh` is linked to `dash` which does not trap the signal and provide its own exit status. Instead, it also dies with `SIGPIPE`.

This is an unfortunate design choice! Because now there is no way to be sure what the exit value of `system` for example will be if there is a signal. It varies depending on what is used for `sh`. GNU cleverness strikes again 😉 .

Oh well. As we've seen in #205 signal exits are kind of an overlooked gray area. I think I will try to avoid spawning a shell at all for this test and just run `yes` directly with an `os/exec.Cmd`.

In my testing I also found some signals are caught by default with shells. For example, I will avoid `SIGQUIT` (3) which is caught and just quits the shell. I used this command to try all the signals:

```
kill -l | ./goawk 'BEGIN{j=1}{for(i=1;i<=NF;i++)printf "%2s %6s %s\n", j++, $i, system("bash -c \047kill -"i" $$\047")}'
```

interp/vm.go
```diff
@@ -1085,17 +1076,23 @@ func (p *interp) callBuiltin(builtinOp compiler.BuiltinOp) error {
 		cmd.Stdout = p.output
 		cmd.Stderr = p.errorOutput
 		_ = p.flushAll() // ensure synchronization
-		err := cmd.Run()
+		err := cmd.Start()
+		if err != nil {
+			// could not start the shell
+			p.printErrorf("%v\n", err)
+			p.replaceTop(num(-1.0))
+			return nil
```
benhoyt, line 1084: Returning will stop executing bytecodes, right? I'm pretty sure we don't want to do that here. Could we see if we can add a test case for this (and then remove the return)?

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+
+import (
+	"bufio"
+	"io"
+	"os/exec"
+	"syscall"
+)
+
+const (
+	NOT_CLOSED_CODE = -127
```
benhoyt, line 11: The capital letter makes this exported as `interp.NOT_CLOSED_CODE`. In addition, in Go people don't normally use `ALL_CAPS` for constants these days. I think we should rename to `notClosedExitCode`.

benhoyt (commented):
I haven't reviewed `iostream.go` or `iostream_test.go` yet (the bulk of the PR!), but here's a few early comments I had time for yesterday. I'll try to get to the rest over the weekend.

Thanks again for this. Note that I'm fairly thorough in my reviews as I like to get quality code merged and code that I can "own" from here on out. Hopefully we'll end up with something we're both proud of!

juster:
> I haven't reviewed `iostream.go` or `iostream_test.go` yet (the bulk of the PR!), but here's a few early comments I had time for yesterday. I'll try to get to the rest over the weekend.
> 
> Thanks again for this. Note that I'm fairly thorough in my reviews as I like to get quality code merged and code that I can "own" from here on out. Hopefully we'll end up with something we're both proud of!

Not a problem and no rush. I think I want to merge the "cleanup" commit into the first commit and pretend it was always cleaned up.

BTW, I found a simple way around the bash problem. I just need to use `exec`. What a lovely, strange command 😄.

interp/vm.go
juster: I seem to remember starting at this. I don't think it will stop _all_ execution, it will just stop the `callBuiltin` function. This is what `callBuiltin` returns at its conclusion.

The goal is to skip the rest of the code in the `case` block. What if I switch the return to `break`? That is less jarring than the `return nil`. I will study the existing interp tests. I seem to remember some test cases there for missing shell commands.

By the way, I split the `exec.Cmd.Run()` into `.Start()` and `.Wait()` so I could separate a starting error and an error result on child process exit. That is because the `cmdExitCode()` assumes the process started.

benhoyt: Ah, sorry -- I thought this was in `execute` instead of `callBuiltin`. No, `return nil` is fine given the context (I was looking at this in GitHub's diff viewer, not my IDE).

Re the separation between `Run` and `Start` -- yep, makes sense, thanks.

interp/iostream_test.go
```diff
@@ -0,0 +1,321 @@
+package interp
```
benhoyt, line 1: I started leaving individual comments on this file, but I'm not sure I agree with the approach you're taking here. These are white-box tests that test internal implementation details. And as noted below, I think some things don't really needs tests at all, like the double-close error.

It seems to me we can test most, if not all, of the expected behaviours by running it through the public `interp` API like the rest of the interpreter tests.

Can we make these regular `interpTests` tests? For example:

```go
var interpTests = []interpTest{
	// ...
	{`BEGIN { print system("kill -1 $$") }`, "", "257\n", "", ""},
}
```

Many of these should even work on Windows, as the default shell there is `sh` too. If not, you can use a special `# !windows-gawk` or `# !windows-goawk` comment -- see [exclude logic](https://github.com/benhoyt/goawk/blob/f2d5f1f1f831a193901126379d9401949d5f6185/interp/interp_test.go#L1005), and we could add another rule so just `# !windows` works too.

There's also the separate [TestShellCommand](https://github.com/benhoyt/goawk/blob/f2d5f1f1f831a193901126379d9401949d5f6185/interp/interp_test.go#L1440) if that's useful to look at.

So overall, let's see how far we can get making these black-box interp tests similar to the existing ones.

interp/iostream_test.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+func checkExitCode(t *testing.T, expectCode, exitCode int) {
+	if exitCode != expectCode {
+		t.Errorf("expected Close() result: %d, got: %d", expectCode, exitCode)
+	}
+}
+
+// Generate a random executable name this is guaranteed not to exist on the $PATH.
+func randMissingExe() string {
```
benhoyt, line 156: Rather than trying to generate a random executable, can't we just use a fixed name that won't exist? In the existing `TestSystemCommandNotFound`, for example, we just use the name `foobar3982`, which seems highly unlikely to exist. :-)

interp/iostream_test.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+// Output checkers are functions which spawn goroutines to read lines from the io.ReaderCloser and
+// check the input. After the content passes or fails the check, they Close() the io.ReadCloser.
+// They send two errors: the first is any problem with the content of the input (or nil) and the second
+// is the error result of Close(). These functions return output checkers as closures.
+
+type outputChecker func(rc io.ReadCloser) chan error
+
+func emptyOutput() outputChecker {
```
benhoyt, line 192: I *think* all of this concurrency/complexity will go away if we can use regular interp tests, as the interpreter handles the command concurrency already.

benhoyt (changes requested):
Thanks @juster! I've done a thorough review now and left a lot of comments. They're mostly fairly minor refactoring requests, but the one major thing is I'd like to restructure the tests to be black-box tests that test external behaviour like the other `interp` tests, rather than testing implementation details. I'm happy to discuss further if that's not possible.

benhoyt:
One other thing: it looks like there's also merge/rebase conflicts from the `ioutil` -> `io` change that I merged -- those will need to be resolved.

interp/interp_test.go
```diff
@@ -1459,7 +1459,7 @@ func TestShellCommand(t *testing.T) {
 			func(config *interp.Config) {
 				config.ShellCommand = []string{"/bin/cat"}
 			})
-		testGoAWK(t, `BEGIN { print system("echo hi") }`, "", "exec: \"foobar3982\": executable file not found in $PATH\n-1\n", "", nil,
+		testGoAWK(t, `BEGIN { print system("echo hi"); print "OK" }`, "", "exec: \"foobar3982\": executable file not found in $PATH\n-1\nOK\n", "", nil,
```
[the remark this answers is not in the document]
juster: It was what you requested earlier to test that the interpreter continues executing after that `return nil` that was added to the `case: BuiltinSystem` line:

https://github.com/juster/goawk/blob/7f1c845517665d6801e0cdf742816e1a559cf30a/interp/vm.go#L1080C1-L1085C4

Do you still want that or no?

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+	"io"
+	"os/exec"
+	"syscall"
+)
+
+const (
+	notClosedExitCode = -127
+)
+
+func cmdExitCode(cmdError error, ioerrs... error) (exitCode int, err error) {
```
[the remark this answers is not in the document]
juster: I think I should really add more comments. This function was part of a refactor to remove the common logic from closing the `exec.Cmd` for both input and output. It is confusing but that is because the _stdlib_ error handling is complicated.

I can merge this function into `exitErrorCode` but I thought it would be less overwhelming to separate them into two phases so that `exitErrorCode` only receives an `*exec.ExitError`. The problem is that `cmdError` might be an `io` error that occurred when closing the `exec.Cmd`. Additionally, there might have been other `io` errors that occurred when flushing the pipe _to_ the `exec.Cmd` before `Wait()`-ing.

On top of that, an `*exec.ExitError` will occur if the child exit `!= 0` but this is already reported by the `ExitCode()`, and so it is discarded. This logic is separated into the `exitErrorCode()` func. That allows any `io` error prior to `Wait()` to bubble up and replace it.

There are two goals here:
1. The command must be `Wait()`-ed on even if an `io` error occurs before the `Wait()` call. Otherwise,  resources are not freed in the `goawk` process for the child.
2. Avoid hiding any errors that occur while fulfilling Goal 1. Some hiding may occur but the "worst" error is always returned.

I should make it more explicit but it is possible for the exit code to indicate success but also return an error code for `io` errors!

### Truth table

Here is a truth table of all possible error conditions. A `Wait()` error that reports a non-zero exit code or signal exit of a child is equal to `nil`.

| Struct | Prior I/O Error | `Wait()` error | Exit Code Result | Error Result |
| ------ | -------------- | ------------- | ---------- | ----- |
| inCmdStream | nil | nil | >= 0 | nil |
| " | nil | IOWaitError | -1 | IOWaitError |
| " | CloseError | nil | >= 0 | CloseError |
| " | CloseError | IOWaitError | -1 | IOWaitError |
| outCmdStream | nil | nil | >= 0 | nil |
| " | nil | IOWaitError | -1 | IOWaitError |
| " | FlushError | nil | >= 0 | FlushError |
| " | FlushError | IOWaitError | -1 | IOWaitError |
| " | CloseError | nil | >= 0 | CloseError |
| " | CloseError | IOWaitError | -1 | IOWaitError |
| " | FlushError + CloseError | nil | >= 0 | FlushError |
| " | FlushError + CloseError | IOWaitError | -1 | IOWaitError |

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+	var ee *exec.ExitError
+	var ok bool
+
+	if cmdError == nil {
+		// exitCode = 0.0
+		goto CheckIOErrors
+	}
+	if ee, ok = cmdError.(*exec.ExitError); !ok {
+		exitCode = -1.0
+		return
```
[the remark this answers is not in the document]
juster: I think this is a typo and it should `return -1, cmdError`. If `ok == false` then the error from `Wait()` is an io-related error.

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+	return
+}
+
+// Copy the behavior of [gm]awk for pipe exit status. n?awk returns the raw
+// wait(3) return value but this is hard to use without a bit shift operator!
+func exitErrorCode(ee *exec.ExitError) (exitCode int, err error) {
+	stat, ok := ee.ProcessState.Sys().(syscall.WaitStatus)
+	if !ok {
+		// Not all platforms support WaitStatus?
+		return -1.0, err
```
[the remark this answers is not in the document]
juster: Yes I think it should return the `err` and just give up if type assertion to `syscall.WaitStatus` fails for some mysterious reason.

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+			break
+		}
+	}
+	return
+}
+
+// Copy the behavior of [gm]awk for pipe exit status. n?awk returns the raw
+// wait(3) return value but this is hard to use without a bit shift operator!
+func exitErrorCode(ee *exec.ExitError) (exitCode int, err error) {
+	stat, ok := ee.ProcessState.Sys().(syscall.WaitStatus)
```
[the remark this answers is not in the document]
juster: Ok

interp/iostream.go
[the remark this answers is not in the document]
juster: Oh I see. It should `return -1, ee`.

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+
+type outputStream interface {
+	io.WriteCloser
+	Flush() error
+	ExitCode() int
+}
+
+type outFileStream struct {
+	*bufio.Writer
+	io.Closer
```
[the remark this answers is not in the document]
juster: If you do that then you would also need to write a `Flush` method one-liner. I think it would be better to leave the `*bufio.Writer` embedded and make an implicit `io.Closer` field. I suppose it doesn't make sense to embed the `io.Closer` if the `Close() method is overridden and that is the only thing the `io.Closer` provides.


```go
type outFileStream struct {
    *bufio.Writer
    closer io.Closer
    exitCode int
}

func (s *outFileStream) Close() error { ... }
```

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+	switch {
+	case err1 != nil:
+		s.exitCode, err = -1, err1
+	case err2 != nil:
+		s.exitCode, err = -1, err2
+	default:
+		s.exitCode = 0
+	}
+	return
+}
```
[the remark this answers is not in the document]
juster: I think this is mostly just a personal aesthetic choice and I don't have a strong opinion about those things. I do see it as a choice between errors, after ensuring that `Flush()` and `Close()` are always run. To me, the suggestion is similar to the `if err := Etc(); err != nil { ... }` pattern which is more appropriate for sequential operations.

I will change it but I prefer the original ofc 😄 .

In all seriousness and not trying to be sarcastic, but if you use the named returns more often you might get used to them and not find them confusing. I am still getting used to them myself. The default values are a little odd sometimes... in other contexts as well... but I'm starting to like them.

Re: your comment on `cmdExitCode` ... right it isn't used here because the extra complexity of `*exec.Cmd` isn't an issue. The reason that `ioerrs` is an `[]error` slice is so that it can be used for consistent error handling for three different callers:

1. `outCmdStream.Close()` has 2 possible prior I/O errors
2. `inCmdStream.Close()` has 1 possible prior I/O error
3. `BuiltinSystem` (from vm.go) has 0 possible prior I/O errors

The slice makes handling all three different situations much simpler.

interp/iostream_test.go
```diff
@@ -0,0 +1,321 @@
...
+	Name, shell string
+	exitCode    int
+	input       string
+	output      outputChecker
+	requires    string
+	skipWin     bool
+}
+
+func TestStreamInFile(t *testing.T) {
+	t.Run("DoubleClose", func(t *testing.T) {
```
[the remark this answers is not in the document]
juster: Well, these are unit tests for the streams themselves. Part of the advantages of the abstracting them is that we can verify the interface between them and the rest of `interp` is consistent.

Yes, right now it's _presumably_ impossible to double close but what if a bug is introduced that causes double-close to happen? Don't you want to know that the expected error conditions are consistent so that it is visible to the rest of the program?

I think the greatest advantage of these lower-level tests are to check that behavior is consistent for all four types of streams. The is running basically the same test for each type of stream which is why I feel a sub-test is appropriate. Sure a flat structure is simpler, but only because it's less work for the writer. I've already taken the trouble to separate them. There is no advantage in mushing them together now.

Coincidentally, a few days ago I added local tests to check the double close for the command streams as well. The `outCmdStream` does not behave as expected and it does not return an error if it is double-closed. If I didn't actually test the error checking I would not have known that?

Maybe if `iostreams` was a separate package you would feel more comfortable unit testing it?

interp/iostream_test.go
```diff
@@ -0,0 +1,321 @@
...
+)
+
+const (
+	shellWaitTime          = 500 * time.Millisecond
+	maxRandExeAttempts     = 10
+	outputStreamBufferSize = 1024
+)
+
+type shellTest struct {
+	Name, shell string
```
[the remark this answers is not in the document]
juster: `Name` is exported so that it can be `testing.T.Run()` as a sub-test.

interp/iostream_test.go
```diff
@@ -0,0 +1,321 @@
...
+	exitCode    int
+	input       string
+	output      outputChecker
+	requires    string
+	skipWin     bool
+}
+
+func TestStreamInFile(t *testing.T) {
+	t.Run("DoubleClose", func(t *testing.T) {
+		f, err := os.CreateTemp("", "iostream_input*")
```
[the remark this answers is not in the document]
juster: or I could just add a `defer f.Remove()`?

interp/vm.go
```diff
@@ -911,33 +910,25 @@ func (p *interp) callBuiltin(builtinOp compiler.BuiltinOp) error {
 		p.replaceTop(num(math.Atan2(y.num(), x.num())))
 
 	case compiler.BuiltinClose:
+		var err error
+		code := float64(-1)
```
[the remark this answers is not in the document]
juster: Ok

interp/iostream_test.go
juster: Ok. I can move as much as possible into `interp_test.go`. But I think the lower-level I/O tests (i.e. "DoubleClose") are still important. I commented on that elsewhere.

I think having separate lower level tests are important for regression. If you have a bug appear it is better to isolate it at as low a level as possible. If a bug occurs in the `iostream` later, it will manifest in the higher level `interpTests` as well as in the stream tests. If the bug occurs at a higher level above the `iostream` it will only manifest in the `interpTests`. This makes it easier to diagnose.

I think your distinction between white box is mostly dependent on `package` boundaries. I am trying to unit test I/O in streams without worrying about the rest of `interp`. This is an attempt to create stratification. Maybe if I moved `iostreams.go` to a separate package it wouldn't be so bothersome to unit test them in isolation?

interp/iostream_test.go
```diff
@@ -0,0 +1,321 @@
...
+		if err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
juster: Ok

interp/iostream_test.go
juster: Ok

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+
+type inCmdStream struct {
+	io.ReadCloser
+	cmd *exec.Cmd
+	exitCode int
+}
+
+func inputCmdStream(cmd *exec.Cmd) (inputStream, error) {
+	rdr, err := cmd.StdoutPipe()
```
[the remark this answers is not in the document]
juster: Ok

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+	ExitCode() int
+}
+
+type outFileStream struct {
+	*bufio.Writer
+	io.Closer
+	exitCode int
+}
+
+func outputFileStream(wc io.WriteCloser, size int) outputStream {
```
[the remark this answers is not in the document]
juster: Ok

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
+package interp
```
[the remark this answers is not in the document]
juster: Ok

interp/interp_test.go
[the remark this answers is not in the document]
benhoyt: Ah, no, I'd just misread what function that `return` was in, I wasn't requesting a test for that. That said, now that I understand the reason for adding this, I'm fine with it -- it certainly doesn't hurt.

interp/iostream.go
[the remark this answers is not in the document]
benhoyt: I'm okay with the switch if you want to leave that, but let's 1) use `flushErr` and `closeErr` instead of `err1` and `err2`, and 2) avoid the naked returns and just do `s.exitCode = -1 \ return flushErr` type of thing.

interp/iostream_test.go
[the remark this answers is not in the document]
benhoyt: Sure, but it doesn't need to be exported, as definition and usage is all in the same package. I tried locally and renaming it to `name` works fine. Though this is probably moot if these tests go away.

interp/iostream_test.go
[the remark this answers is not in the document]
benhoyt: I'm okay leaving a couple of specific tests like the double-close one if you think it's going to catch bugs. However, in the case of the double-close, I'm still not sure it's worth it: why do we need to assert that it returns an error if double-closed when we're not double-closing it in the code?

In any case, I strongly believe we should test as much as possible using interp or similar white-box tests. These have proved very valuable in the past, for a couple of reasons:

1) I've been able to do large refactorings of the interpreter without the test suite changing one bit. As a result, I have a ton of confidence that the refactorings haven't broken anything. On the flip hand, if the tests are tied to the implementation, you're testing that specific implementation, so when the implementation changes, the tests break (and you don't know if that's because the old implementation was bad, or just because the tests relied on implementation details -- probably the latter). So white-box tests are resilient in the face of implementation changes.

2) All the interp tests are also run through Gawk. This ensures I've got them right and that Gawk has the same behaviour. A small number of tests use `# !gawk` type comments to not run on gawk, but almost all of them are run on both Gawk and goawk. The has helped me numerous times in the past. Similarly, I run the entire nawk and gawk test suites on Goawk as well, and this is of course done white-box, by feeding source code to GoAWK, sending input, and checking output.

Once again, I'm okay with leaving a couple of specific "programmer error" tests in `iostream_test.go`, but let's redo the rest as interp tests. (I don't think we should make `iostreams` a separate package.)

interp/iostream_test.go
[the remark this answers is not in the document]
benhoyt: That works too, though I'd say `t.TempDir()` is specifically designed for this kind of use case, and has the nice property of cleaning up after itself.

interp/iostream_test.go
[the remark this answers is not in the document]
benhoyt: Oops, I said "white box" above when I meant "black box". Updating.

interp/iostream_test.go
benhoyt: See my comment above as well, but in short: I'm okay with a couple of "programmer error" tests that test implementation details, but yeah, let's move the rest to interp tests.

I think the occasional programmer error test is okay to be white box, but in general I disagree somewhat with you here: I think testing top-level contracts is much more useful and powerful than testing implementation details. Doing so allows you to refactor the internals majorly and have your test suite just keep on telling you what you need to know. You're right that sometimes lower-level tests allow you do diagnose a bug more quickly (closer to its source), but that's fairly rare.

I don't think it has much to do with package boundaries: what I'm suggesting is we test (as much as possible) the contract of essentially the entire program against the spec (POSIX spec or Gawk or whatever). That's what has allowed me to run these tests against multiple implementations (GoAWK, Gawk, nawk).

interp/iostream.go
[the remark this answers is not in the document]
benhoyt: Thanks for the effort on your comment and the truth table. However, I don't think "more comments" or a big truth table will really help here -- and it may just make it more text to wade through and so harder to understand. I think the reason it's hard to follow now is because you're trying to abstract too much: handle three different cases, with multiple errors as input, and then the `goto` and bare returns don't help clarity.

So here's a concrete suggestion that I think will work and (to me) is very clear and simple: How about we move only the `exec.ExitError` check into `exitErrorCode`, and have it take a regular `error`, like so:

```go
func exitErrorCode(cmdError error) (exitCode int, err error) {
	if ee, ok = cmdError.(*exec.ExitError); !ok {
		return -1, err
	}
	// ... rest of exitErrorCode
}
```

And then make the three cases that use it concrete, boring, and clear by just spelling out the errors they check. Like so:

```go
// system()
err = cmd.Wait()
if err != nil {
	// ... check ctx
	code, err := exitErrorCode(err)
	ret = float64(code)
	if err != nil {
		p.printErrorf("%v\n", err)
	}
}

func (s *outCmdStream) Close() (err error) {
	flushErr := s.Writer.Flush()
	closeErr := s.Closer.Close()
	cmdErr := s.cmd.Wait()
	code, err := exitErrorCode(cmdErr)
	s.exitCode = code
	if err != nil {
		return err
	}
	if flushErr != nil {
		return flushErr
	}
	if closeErr != nil { // could just "return closeErr" here, but I think this is slightly clearer
		return closeErr
	}
	return nil
}

func (s *inCmdStream) Close() error {
	closeErr := s.ReadCloser.Close()
	cmdErr := s.cmd.Wait()
	code, err := exitErrorCode(cmdErr)
	s.exitCode = code
	if err != nil {
		return err
	}
	if closeErr != nil {
		return closeErr
	}
	return nil
}
```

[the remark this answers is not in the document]
juster: Oh ok I see what you are saying now. That sounds good to simplify.

interp/iostream_test.go
juster: I haven't had much time this week but I was working out where to put the tests in interp_test.

benhoyt: No problem. I was thinking it'd be good if as many of them as possible could go in the list of generic interp tests, see the [`interp_tests`](https://github.com/benhoyt/goawk/blob/f2d5f1f1f831a193901126379d9401949d5f6185/interp/interp_test.go#L43) slice.

interp/iostream_test.go
```diff
@@ -0,0 +1,321 @@
...
+		in := inputFileStream(f)
+		f.Close()
+		if err := in.Close(); err == nil {
+			t.Error("expected inputStream.Close to error")
+		}
+		checkExitCode(t, -1.0, in.ExitCode())
+	})
+}
+
+func TestStreamOutFile(t *testing.T) {
```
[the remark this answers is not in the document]
juster: Ok

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+// wait(3) return value but this is hard to use without a bit shift operator!
+func exitErrorCode(ee *exec.ExitError) (exitCode int, err error) {
+	stat, ok := ee.ProcessState.Sys().(syscall.WaitStatus)
+	if !ok {
+		// Not all platforms support WaitStatus?
+		return -1.0, err
+	}
+	switch {
+	case stat.Exited():
+		exitCode = stat.ExitStatus()
```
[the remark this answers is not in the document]
juster: Ok

interp/iostream_test.go
juster: I've moved everything other than the test for double-close conditions to `interp_test.go`. This was a problem for `outCmdStream`-s because they did not detect the double close as error and overwrote the `exitCode`.

interp/iostream_test.go
juster: I've removed this test completely.

interp/iostream_test.go
juster: Ok, removed.

interp/iostream_test.go
juster: Also, I really don't think Windows supports POSIX signals but you're welcome to try if you have a Windows box handy. IIRC, cygwin offers some sort of emulation but I wouldn't assume go on Windows is compatible with this.

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+func (s *outNullStream) Close() error {
+	return nil
+}
+
+func (s *outNullStream) ExitCode() int {
+	return -1
+}
+
+type inFileStream struct {
+	io.ReadCloser
```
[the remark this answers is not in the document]
juster: I don't feel that would be an improvement. Creating a separate `closer` field in `outCmdStream` made sense because we were not using embedding to pass through the `Close()` method to the `io.Closer`. The `io.Closer` of course only offers the `Close()` method so it was pointless to embed it. That is not the case here.

interp/iostream.go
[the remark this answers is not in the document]
juster: I've uplifted the error checking into a separate function:

```go
// firstError returns the first non-nill error or nil if all errors are nil.
func firstError(errs ...error) error {
    for _, err := range errs {
        if err != nil {
            return err
        }
    }
    return nil
}
```

Now it is no longer necessary to pass in the previous errors to the `cmdExitCode` function and it is much simpler (i.e. no `goto` statements). I've also changed the `cmdExitCode` function to `waitExitCode` and perform the `.Wait()` inside of it, because a `.Wait()` always preceded calling `cmdExitCode`.

Now I can use `firstError` in three places of `iostream.go` and avoid using it for the `system()` builtin function. I think it makes the error priority much more obvious, on a single line, compared to reading through 10 lines of boilerplate:

```go
	s.exitCode, waitErr = waitExitCode(cmd)
	return firstError(cmdErr, flushErr, closeErr)
	// VS
	if cmdErr != nil {
		return cmdErr
	}
	if flushErr != nil {
		return flushErr
	}
	if closeErr != nil {
		return closeErr
	}
	return nil
```

interp/iostream.go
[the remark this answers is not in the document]
juster: I've removed the named returns and there is now a check for double-closing in every Close() method:
```go
func (s *outFileStream) Close() error {
	if s.exitCode != notClosedExitCode {
		return doubleCloseError
	}
	flushErr := s.Writer.Flush()
	closeErr := s.closer.Close()
	if err := firstError(flushErr, closeErr); err != nil {
		s.exitCode = -1.0
		return err
	}
	s.exitCode = 0.0
	return nil
}
```

interp/iostream.go
```diff
@@ -0,0 +1,203 @@
...
+		w.Close()
+		return nil, err
+	}
+	out := &outCmdStream{w, bufio.NewWriterSize(w, outputBufSize), cmd, notClosedExitCode}
+	return out, nil
+}
+
+func (s *outCmdStream) Close() (err error) {
+	err1 := s.Writer.Flush()
+	err2 := s.Closer.Close()
```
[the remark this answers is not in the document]
juster: Ok, these have been renamed.

interp/iostream_test.go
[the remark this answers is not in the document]
juster: This was removed.

That's what I thought too, I must have been confused by error messages when I was switching it back and forth in a hurry.

interp/iostream_test.go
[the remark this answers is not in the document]
juster: I've left a single `TestStreamDoubleClose` test inside of `iostream_test.go` and it has the 4 combinations of input/output file/pipe as subtests.

interp/interp_test.go
```diff
@@ -840,6 +841,13 @@ BEGIN { x[1]=3; f5(x); print x[1] }
 	{`BEGIN { print >"out"; close("out"); getline <"out"; print >"out" }  # !awk !gawk`, "", "", "can't write to reader stream", ""},
 	{`BEGIN { print >"out"; close("out"); getline <"out"; print |"out" }  # !awk !gawk`, "", "", "can't write to reader stream", ""},
 
+	// The value of close() on a pipe emulates gawk behavior. Results are identical for both
+	// input and output. Windows does not do POSIX signals.
+	{`BEGIN { cmd="exit 9"; print "" |cmd; print close(cmd) } # !awk !posix`, "", "9\n", "", ""},
+	{`BEGIN { cmd="exec kill -9 $$"; print "" |cmd; print close(cmd) } # !awk !posix !windows-gawk`, "", "265\n", "", ""},
+	{`BEGIN { cmd="exit 9"; cmd |getline; print close(cmd) } # !awk !posix`, "", "9\n", "", ""},
+	{`BEGIN { cmd="exec kill -9 $$"; cmd |getline; print close(cmd) } # !awk !posix !windows-gawk`, "", "265\n", "", ""},
+
```
benhoyt, line 850: The Windows tests are currently failing. I think we'll need to skip the `kill -9` ones on Windows (not just `windows-gawk`). Perhaps add `!windows` and then add this to the `TestInterp` runner:

```go
if strings.Contains(test.src, "!windows") && runtime.GOOS == "windows" {
    t.Skip("skipping on windows")
}
```

interp/iostream_test.go
```diff
@@ -0,0 +1,82 @@
...
+	exitCode := in.ExitCode()
+	if err := in.Close(); !errors.Is(err, doubleCloseError) {
+		t.Error("expected inputStream.Close() to return error on double close")
+	}
+	if in.ExitCode() != exitCode {
+		t.Error("expected inputStream.ExitCode() to stay the same")
+	}
+}
+
+func checkDoubleOutputClose(t *testing.T, out outputStream) {
```
benhoyt, line 71: We should probably mark these 2 check functions with `t.Helper()` so the stack frame / line number shows as expected in case of failure.

interp/iostream_test.go
```diff
@@ -0,0 +1,82 @@
...
+		out, err := newOutCmdStream(cmd)
+		if err != nil {
+			t.Fatal(err)
+		}
+		checkDoubleOutputClose(t, out)
+	})
+}
+
+func execDefShell(scriptlet string) *exec.Cmd {
+	cmdline := append(getDefaultShellCommand(), scriptlet)
```
benhoyt, line 54:
```suggestion
	cmdline := append(defaultShellCommand, scriptlet)
```

interp/iostream_test.go
benhoyt: Actually, it looks like these functions are identical apart from the type. Could add

```go
type inOutCloser {
    Close() error
    ExitCode() float64
}
```

And then combine them into the same function (both `inputStream` and `outputStream` implement that). Up to you.

interp/iostream.go
```diff
@@ -0,0 +1,209 @@
...
+import (
+	"bufio"
+	"errors"
+	"io"
+	"os/exec"
+	"syscall"
+)
+
+const (
+	notClosedExitCode = -127.0
```
benhoyt, line 15: This is weird as a float. Let's make it -127 integer (shouldn't matter as an untyped constant anyway).

Edit: actually, I think we should avoid this altogether -- see suggestion below.

interp/iostream.go
```diff
@@ -0,0 +1,209 @@
...
+)
+
+var (
+	doubleCloseError = errors.New("close: stream already closed")
+)
+
+// firstError returns the first non-nill error or nil if all errors are nil.
```
benhoyt, line 22:
```suggestion
// firstError returns the first non-nil error or nil if all errors are nil.
```

interp/iostream.go
```diff
@@ -0,0 +1,209 @@
...
+	case status.Exited():
+		return float64(status.ExitStatus()), nil
+	default:
+		return -1.0, err
+	}
+}
+
+type inputStream interface {
+	io.ReadCloser
+	ExitCode() float64
```
benhoyt, line 69: I think it's a bit weird that these low-level interfaces are dealing in float64s (that's a higher-level AWK value thing). Let's change `iostream.go` to deal in `int` exit codes, and only convert to `float64` where needed in the callers (even if it means a few more type conversions there).

interp/iostream.go
```diff
@@ -0,0 +1,209 @@
...
+	exitCode float64
+}
+
+func newOutFileStream(wc io.WriteCloser, size int) outputStream {
+	b := bufio.NewWriterSize(wc, size)
+	return &outFileStream{b, wc, notClosedExitCode}
+}
+
+func (s *outFileStream) Close() error {
+	if s.exitCode != notClosedExitCode {
```
benhoyt, line 90: I would prefer this were a separate `closed bool` on the type, rather than trying to stuff it into the `exitCode` field. That would avoid the magic `notClosedExitCode` constant, and make this test:

```go
if s.closed {
    return doubleCloseError
}
s.closed = true
```

interp/iostream.go
```diff
@@ -0,0 +1,209 @@
...
+		return doubleCloseError
+	}
+	flushErr := s.Writer.Flush()
+	closeErr := s.closer.Close()
+	if err := firstError(flushErr, closeErr); err != nil {
```
benhoyt, line 95: Yep, this is nice and readable -- thanks for the update.

benhoyt (changes requested):
Excellent, thanks! Just a few very minor change requests now.

interp/iostream.go
juster: Ok. It seems redundant to me but that's fine. Notice that the only way it could be a negative number, or that specific exit code, is if `exitCode` is initialized with that and never overwritten. But sure.

interp/iostream.go
juster: Ok sure. But keep in mind the exit codes themselves are also the expected higher-level awk values.

interp/interp_test.go
juster: Ok.

It looks like the cygwin signal emulation (or maybe mingw?) is automatically changing the exit status to `signal << 8`. A pity really. Because there technically are no signals in Windows, I believe the `syscall.WaitStatuc.Signal()` method will never return `true`.

See: `GOOS=windows go doc -src syscall.WaitStatus.Signaled`.

interp/iostream_test.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+func TestStreamDoubleClose(t *testing.T) {
+	dir := t.TempDir()
+	t.Run("InFile", func(t *testing.T) {
+		f, err := os.Create(filepath.Join(dir, "infile"))
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer os.Remove(f.Name())
```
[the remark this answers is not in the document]
juster: Oh forgot those.

interp/iostream_test.go
juster: Ok.

interp/iostream_test.go
juster: Ok.

interp/interp_test.go
juster: I've added the `!windows` flag but I assume it means that the _goawk_ test should be skipped on Windows and not that all tests should be skipped on Windows. Otherwise the `!windows-gawk` flag becomes redundant and also, checking for `!windows` with contains will return true for `!windows-gawk` as well.

Because `strings.Contains(test.src, "!windows")` will also match `!windows-gawk` erronously (I assume), I have implemented a simple tokenizer that will check for flags after the comment and split them up, storing them into a map of bools, etc.

This should prove something interesting which I found when I dusted off my windows box and run these: gawk successfully returns the proper `close()` value when a child process is killed via signal. I suspect this is because the chocolatey gawk is compiled (statically) with cygwin. This is hard to prove as the package repo only hosts binaries. IOW the gawk test will pass... but not with --posix ofc ;-).

I also found and fixed a small number of tests which had `!gawk` flags but no comment char.

juster: According to this: https://github.com/benhoyt/goawk/actions/runs/6213666312/job/16864921268?pr=206 ...

_gawk_ from _chocolatey_ is running `cmd.exe` and not `sh` like you do for _goawk_ on Windows.

When I run the tests with _cygwin_ on Windows, where _gawk_ is installed by _cygwin_, these tests pass. That means _gawk_ on _cygwin_ is running piped commands in `sh` and not `cmd.exe`. (Other tests fail though.)

I suppose I should just skip all the signal tests on Windows by using `!windows !windows-gawk`?

Honestly this compatibility testing is a bit of a pain. I can see it as being useful in a generated report but embedding it into the CI as a go test for your own implementation is asking for trouble.

juster: I've added `!windows-gawk` and replaced `kill` with `/bin/kill` since it is more accurate, if we are not testing Windows anyways.

interp/interp_test.go
```diff
@@ -587,13 +587,14 @@ BEGIN {
...
-	{`BEGIN { print match("food", "foo"), RSTART, RLENGTH }  !gawk`, "", "1 1 3\n", "", ""},
-	{`BEGIN { print match("x food y", "fo"), RSTART, RLENGTH }  !gawk`, "", "3 3 2\n", "", ""},
-	{`BEGIN { print match("x food y", "fox"), RSTART, RLENGTH }  !gawk`, "", "0 0 -1\n", "", ""},
-	{`BEGIN { print match("x food y", /[fod]+/), RSTART, RLENGTH }  !gawk`, "", "3 3 4\n", "", ""},
-	{`BEGIN { print match("絵 fööd y", /[föd]+/), RSTART, RLENGTH }  !gawk`, "", "5 5 6\n", "", ""},
+	{`BEGIN { print match("food", "foo"), RSTART, RLENGTH } # !gawk`, "", "1 1 3\n", "", ""},
+	{`BEGIN { print match("x food y", "fo"), RSTART, RLENGTH } # !gawk`, "", "3 3 2\n", "", ""},
+	{`BEGIN { print match("x food y", "fox"), RSTART, RLENGTH } # !gawk`, "", "0 0 -1\n", "", ""},
+	{`BEGIN { print match("x food y", /[fod]+/), RSTART, RLENGTH } # !gawk`, "", "3 3 4\n", "", ""},
+	{`BEGIN { print match("絵 fööd y", /[föd]+/), RSTART, RLENGTH } # !gawk`, "", "5 5 6\n", "", ""},
```
benhoyt, line 597: Ah nice -- thanks for spotting these.

interp/interp_test.go
benhoyt: No, I was definitely thinking `!windows` should mean skip this test on Windows, regardless of which interpreter (goawk, gawk, gawk --posix). I think `!windows !windows-gawk` is redundant and odd. So let's change `!windows` so it applies to both.

So this will mean we can skip all the signal tests on Windows by using just `!windows`.

The little tokenizer is a nice addition, though -- thanks!

Regarding this compatibility testing being a bit of pain -- you're not wrong, it's a bit messy. However, I think cleaning it up further is outside the scope of this PR, so let's just leave it for now.

interp/interp_test.go
```diff
@@ -983,6 +991,25 @@ BEGIN { x[1]=3; f5(x); print x[1] }
 	{`{ print $1<$2 }`, "1_0 2", "1\n", "", ""},
 }
 
+func (it interpTest) Flags() map[string]bool {
```
benhoyt, line 994: The fact that this map included both true and false items was a bit unintuitive to me at first. Perhaps add a doc comment like:

```suggestion
// Flags parses the "skip flags" in it.src's first comment and returns them as
// a map. An enable flag ("# +flag") appears as {"flag": true} in the map, and
// a disable flag ("# !flag") appears as {"flag": false} in the map.
func (it interpTest) Flags() map[string]bool {
```

interp/interp_test.go
```diff
@@ -983,6 +991,25 @@ BEGIN { x[1]=3; f5(x); print x[1] }
...
+	flags := make(map[string]bool)
+	i := strings.Index(it.src, "#")
+	if i < 0 || i+1 == len(it.src) {
```
benhoyt, line 997: The second clause isn't needed, as `strings.Fields(it.src[i+1:])` will return an empty slice in that case:

```suggestion
	if i < 0 {
```

interp/interp_test.go
```diff
@@ -1055,6 +1084,9 @@ func TestInterp(t *testing.T) {
 
 		// Then test it in GoAWK
 		t.Run(testName, func(t *testing.T) {
+			if f, ok := flags["windows"]; !f && ok && runtime.GOOS == "windows" {
+				t.Skipf("skipping windows")
+			}
```
benhoyt, line 1089: Per comment above, we'll need to include this logic in the `runAWK` section above too.

interp/iostream.go
```diff
@@ -0,0 +1,224 @@
...
+
+func newOutNullStream() outputStream { return &outNullStream{io.Discard, false} }
+func (s outNullStream) Flush() error { return nil }
+func (s *outNullStream) Close() error {
+	if s.closed {
+		return doubleCloseError
+	}
+	s.closed = true
+	return nil
+}
```
benhoyt, line 161: I'm not sure this is necessary for `outNullStream` -- double-close it *not* an error for that type, right? But okay to leave it for consistency if you prefer.

interp/iostream_test.go
```diff
@@ -0,0 +1,73 @@
...
+		cmd := execDefShell("echo close me")
+		out, err := newOutCmdStream(cmd)
+		if err != nil {
+			t.Fatal(err)
+		}
+		checkDoubleClose(t, out)
+	})
+}
+
+func execDefShell(scriptlet string) *exec.Cmd {
```
benhoyt, line 51: "def" sounds more like "define" to me in a programming context. Let's just spell this out:

```suggestion
func execDefaultShell(scriptlet string) *exec.Cmd {
```

interp/interp_test.go
```diff
@@ -983,6 +991,25 @@ BEGIN { x[1]=3; f5(x); print x[1] }
 	{`{ print $1<$2 }`, "1_0 2", "1\n", "", ""},
 }
 
+func (it interpTest) Flags() map[string]bool {
+	flags := make(map[string]bool)
+	i := strings.Index(it.src, "#")
+	if i < 0 || i+1 == len(it.src) {
+		return flags
+	}
+	for _, fld := range strings.Fields(it.src[i+1:]) {
```
benhoyt, line 1000: Naming nit -- no reason to avoid vowels here:

```suggestion
	for _, field := range strings.Fields(it.src[i+1:]) {
```

benhoyt (changes requested):
Thanks for the updates and the little flags parser! Just one more change requested (and a couple of nits): I was intending that `!windows` mean "skip on windows, regardless of interpreter" as I think that makes the most sense (and fits with the meaning of Go's own `!windows` build constraint).

interp/iostream_test.go
juster: The verb already exists so it can't be "define". That would be "execute define shell" which is meaningless.

As you wish.

interp/iostream.go
juster: This is necessary to prevent the `outputStream` from being a leaky abstraction. The point is to produce the same behavior no matter which underlying implementation is used for the interface.

interp/interp_test.go
juster: Ok

interp/interp_test.go
juster: Ok

interp/interp_test.go
juster: As you wish.

benhoyt:
Thanks for your efforts here -- merged! I intend to address a couple of other issues and then release a new GoAWK version that includes this change.

juster:
Sweet thanks!

I noticed this failed test: https://github.com/benhoyt/goawk/actions/runs/6293036690/job/17083040381

It looks like there is a race and the shell exited before goawk could write to the pipe. I should probably change `exit 9` to `read; exit 9` to force the shell to read the empty line before exiting.

benhoyt:
Sounds good -- please put up another PR with the fix. For the record, I can get it to fail a few times locally by running it 1000 times:

```
go test -v -count=1000 -run='^TestInterp/^BEGIN' ./interp
```

PR #209, opened 2023-09-10
flushAll() stops flushing on first error

### LINKED ISSUE #208: fflush() stops on first flush error
During my reading of the `interp.flushAll` method I noticed that it accidentally uses conditional evaluation. Because the `&&` operator is "short-circuited", if any `Flush()` returns a non-nil error, then all of the following `Flush()`-es will be skipped. `allGood` will be false so the right-hand side of `&&` will not be evaluated.

```go
// Flush all output streams as well as standard output. Report whether all
// streams were flushed successfully (logging error(s) if not).
func (p *interp) flushAll() bool {
	allGood := true
	for name, writer := range p.outputStreams {
		allGood = allGood && p.flushWriter(name, writer)
	}
	if _, ok := p.output.(flusher); ok {
		// User-provided output may or may not be flushable
		allGood = allGood && p.flushWriter("stdout", p.output)
	}
	return allGood
}
```

Code changes are trivial but I'm not sure how to create a regression test as an interpreter "black box" test.

juster:
The && operator causes conditional evaluation (aka "short-circuiting"). This means if the left-hand-side is false the right hand side is not evaluated. If any call to interp.flushWriter returns false then all streams following that stream in the range, as well as interp.output, will not be flushed.

Removes the type cast to the flusher interface because that is done in interp.flushWriter. This check would be performed twice for any interp.output value whose type implements the interface.

Fixes #208

### DIFF
--- a/interp/io.go
+++ b/interp/io.go
@@ -865,11 +865,12 @@ func (p *interp) closeAll() {
 func (p *interp) flushAll() bool {
 	allGood := true
 	for name, writer := range p.outputStreams {
-		allGood = allGood && p.flushWriter(name, writer)
+		if !p.flushWriter(name, writer) {
+			allGood = false
+		}
 	}
-	if _, ok := p.output.(flusher); ok {
-		// User-provided output may or may not be flushable
-		allGood = allGood && p.flushWriter("stdout", p.output)
+	if !p.flushWriter("stdout", p.output) {
+		allGood = false
 	}
 	return allGood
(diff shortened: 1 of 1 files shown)


juster:
Let me know if you want me to add regression tests.

benhoyt:
Good catch. This looks excellent. I think it's tough to test (even with a white-box test), so let's just leave as is.

PR #213, opened 2023-10-08
Fix rare race in pipe-to-command close() result test

juster:
Here a fix for the test race condition in #206 and the test fail for Linux:

https://github.com/benhoyt/goawk/actions/runs/6293036690/job/17083040381.

I was able to reproduce on my older Macbook and Go 1.17 using `go test -count 1000 -run='TestInterp/^BEGIN { cmd'`... but not consistently. In my case the other pipe-to-command test failed (the one which uses `/bin/kill`) but with the same pipe error.

Analysis/commit message follows:

The bufio.Writer does not write to the Cmd pipe until Flush is called when the Cmd is Wait-ed on. This causes a Flush error but not a Wait error. This only happens if the Cmd starts and exits before the Flush can be called by goawk.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -843,9 +843,9 @@ BEGIN { x[1]=3; f5(x); print x[1] }
 	{`BEGIN { print >"out"; close("out"); getline <"out"; print |"out" }  # !awk !gawk`, "", "", "can't write to reader stream", ""},
 
 	// The value of close() on a pipe emulates gawk behavior. Results are identical for both
-	// input and output. Windows does not do POSIX signals.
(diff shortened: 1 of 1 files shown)


juster:
I fixed a Linux test failure which was caused by using sh's (technically bash's) `read` with no arguments. But Windows is a puzzle to me, since it has no `read` builtin at all. I'm out of time tonight I'll have to come back to this later.

juster:
I looked at this today and it was more obvious what the problem was. The `gawk` installed with `chocolatey` on the `windows` runner does not run command pipes using `sh.exe`. Instead it runs commands with `cmd.exe`.

I skipped the test which uses the `read` shell command by using `!windows-gawk`.

benhoyt (approved):

benhoyt:
Thanks!

PR #214, opened 2023-10-09
Fix #190: FS regex [^,]* fails on empty field

### LINKED ISSUE #190: FS regex [^,]* fails on empty field
For example:

```
$ echo 'x,,z' | goawk -F'[^,]*' '{for (i=1;i<=NF;i++) print i, "<"$i">"}'
1 <>
2 <,>
3 <,>
4 <>

# Should be (and other AWKs print):
$ echo 'x,,z' | gawk -F'[^,]*' '{for (i=1;i<=NF;i++) print i, "<"$i">"}'
1 <>
2 <,,>
3 <>
```

See https://www.austingroupbugs.net/view.php?id=1468 (0001468).

Thanks @raygard for the report.

fioriandrea:
This pull request fixes issue #190 by introducing a custom split function that skips empty matches (according to https://www.austingroupbugs.net/view.php?id=1468, this is the correct behaviour). It is used only for FS

Fixes #190

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -65,6 +65,12 @@ NR==2, NR==4 { print $0 }
 NR==3, NR==5 { print NR }
 `, "a\nb\nc\nd\ne\nf\ng", "b\nc\n3\nd\n4\n5\n", "", ""},
 
+	// field separator cannot match empty field
+	{`BEGIN {FS = "[^,]*"} {for (i=1;i<=NF;i++) print i, "<"$i">"}`, `x,,z`, "1 <>\n2 <,,>\n3 <>\n", "", ""},
+	{`BEGIN {FS = "[^a-z]*"} {for (i=1;i<=NF;i++) print i, "<"$i">"}`, `xyz`, "1 <xyz>\n", "", ""},
+	{`BEGIN {FS = "[0-9]*"} {for (i=1;i<=NF;i++) print i, "<"$i">"}`, `xyz`, "1 <xyz>\n", "", ""},
+	{`BEGIN {FS = "(abc)?"} {for (i=1;i<=NF;i++) print i, "<"$i">"}`, `abcdef`, "1 <>\n2 <def>\n", "", ""},
+
 	// print and printf statements
 	{`BEGIN { print "x", "y" }`, "", "x y\n", "", ""},
 	{`BEGIN { print OFS; OFS = ","; print "x", "y" }`, "", " \nx,y\n", "", ""},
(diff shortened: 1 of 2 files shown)


benhoyt (commented):
Thanks for this!

I'll review this in detail later, but for now, can we please have a couple of test cases in `interp_test.go`?

Also, it'd be nice to see a small benchmark (not an official `BenchmarkFoo` func, but just one you run manually on a semi-large input with custom matching -- run it before and after and post the numbers, so ensure we haven't regressed in speed much.

fioriandrea:
The results of a benchmark I ran manually are the following:

```sh
ls -lh ../data.bench 
-rw-r--r-- 1 af af 353M 10 ott 19.30 ../data.bench

# with custom split
time ./goawk -F '[0123456789!@#$%^&*()_+]+' '{ print $20 }' ../data.bench >/dev/null

real    0m27,727s
user    0m28,734s
sys     0m0,536s

# without custom split
time ./goawk -F '[0123456789!@#$%^&*()_+]+' '{ print $20 }' ../data.bench >/dev/null

real    0m28,043s
user    0m28,245s
sys     0m0,186s
```

I generated the benchmark data with the following script:

```awk
#/bin/awk -f

BEGIN {
    columns = 10000
    rows = 4000
    maxDelimLen = 10
    len = split("0123456789!@#$%^&*()_+", chars, "")
    for (r = 1; r <= rows; r++) {
        for (c = 1; c <= columns; c++) {
            delimLen = int(maxDelimLen * rand()) + 1
            printf("ABCD")
            for (i = 1; i <= delimLen; i++) {
                char = chars[int(len * rand())]
                printf(char)
            }
        }
        printf("ABCD")
        printf("\n")
    }
}
```

Just for curiosity, I tried to run the same benchmark on gawk, and I got the following results:

```sh
time gawk -F '[0123456789!@#$%^&*()_+]+' '{ print $20 }' ../data.bench >/dev/null

real    0m0,228s
user    0m0,183s
sys     0m0,045s
```

I suspect that gawk doesn't split fields if it doesn't need to. In contrast, in goawk all the fields are generated for each line (in this case, split has been called 4000 times, generating a slice containing all fields each time. This is done both with my custom split and with the built-in split).

benhoyt:
Thanks for this! I made a small tweak to allow reuse of a pre-allocated `p.fields` slice. I'm satisified that at least it's not any/much slower than what we what. Wonder what Gawk is doing there, though! Go's regexp package is fairly slow, but I wouldn't have thought that slow... still, that's a crazy input dataset you have there. :-)

PR #216, opened 2023-10-21
Fix wrong precedence of 'expr | getline' expressions

### LINKED ISSUE #189: Parsing "expr | getline > 0" fails without extra parens
When parsing a getline expression like `expr | getline > 0`, GoAWK fails to parse it where other AWKs succeed. For example:

```
$ goawk 'BEGIN {while ("ls README*" | getline > 0) print}'
<cmdline>:1:38: expected ) instead of >
BEGIN {while ("ls README*" | getline > 0) print}
                                     ^

# Gawk (and others) succeed:
$ gawk 'BEGIN {while ("ls README*" | getline > 0) print}'
README.md

# GoAWK needs extra parentheses
$ goawk 'BEGIN {while (("ls README*" | getline) > 0) print}'
README.md
```

Thanks @raygard for the bug report.

fioriandrea:
This pull request modifies `getLine()` inside parser.go so that it lets binary operators come after 'expr | getline' expressions.

I also uncommented two tests that were in TODO inside interp_test.go and added two new tests.

There were two other tests under the same TODO: (`BEGIN { print("echo foo" | getline ($0+1)); print }` and `BEGIN { print("echo foo" | getline foo()); print } function foo() { print "z" }`), but those address a separate issue.

Fixes #189

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -124,9 +124,10 @@ type parser struct {
 	val     string   // string value of last token (or "")
 
 	// Parsing state
-	inAction  bool   // true if parsing an action (false in BEGIN or END)
-	funcName  string // function name if parsing a func, else ""
-	loopDepth int    // current loop depth (0 if not in any loops)
+	inAction           bool   // true if parsing an action (false in BEGIN or END)
+	funcName           string // function name if parsing a func, else ""
+	loopDepth          int    // current loop depth (0 if not in any loops)
+	pendingGetLineLeft ast.Expr
 
 	// Variable tracking and resolving
 	multiExprs map[*ast.MultiExpr]Position // tracks comma-separated expressions
@@ -518,21 +519,20 @@ func (p *parser) exprList(parse func() ast.Expr) []ast.Expr {
 // which skips PIPE GETLINE and GREATER expressions.
 
 // Parse a single expression.
-func (p *parser) expr() ast.Expr      { return p.getLine() }
+func (p *parser) expr() ast.Expr      { return p._assign(p.getLine) }
 func (p *parser) printExpr() ast.Expr { return p._assign(p.printCond) }
 
 // Parse an "expr | getline [lvalue]" expression:
 //
 //	assign [PIPE GETLINE [lvalue]]
 func (p *parser) getLine() ast.Expr {
-	expr := p._assign(p.cond)
+	p.pendingGetLineLeft = nil
+	left := p.cond()
 	if p.tok == PIPE {
-		p.next()
-		p.expect(GETLINE)
-		target := p.optionalLValue()
-		return &ast.GetlineExpr{expr, target, nil}
+		p.pendingGetLineLeft = left
+		return p.cond()
 	}
-	return expr
+	return left
 }
 
 // Parse an = assignment expression:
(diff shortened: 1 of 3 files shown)


benhoyt:
Hmm, now that I see this and have a play with it, I'm not sure about this. It seems to me a bit too ad-hoc and special-casey. Perhaps that's part of the reason it doesn't yet work with the two commented-out tests (and with the Gawk `getline` and `getline3` tests skipped in `goawk_test.go`). What are we missing that would make this more general?

fioriandrea:
@benhoyt I reverted the changes I made with the `matches()` function and renamed the PR. After more carefully thinking about the problem, I realized that:

- before `|`, `getline` has less precedence than all expressions except assignments
- after `|`, `getline` has the same precedence as a primary expression

Indeed, `"echo " "date" | getline + 1` gets parsed as `(("echo " "date") | getline) + 1` by all other awk implementations.

Although the code I just committed is a bit hacky, it fixes all broken test cases. Basically, I save the lhs of the `expr | getline` expression (in the example above,  `"echo " "date"`) in the parser structure and then I handle the pipe operator inside the `primary` function, where I finish building the `expr | getline` expression. This is done to be in line with the behavior I described in the bullet list above.

I would like to hear your opinion on this (especially if you can think of a less hacky way to handle this)

parser/parser.go
```diff
@@ -807,6 +808,16 @@ func (p *parser) primary() ast.Expr {
 			file = p.primary()
 		}
 		return &ast.GetlineExpr{nil, target, file}
+	case PIPE:
+		if p.pendingGetlineLeft == nil {
+			panic(p.errorf("expected expression instead of %s", p.tok))
```
benhoyt, line 813: Is this possible to hit? What AWK source code would cause it?

benhoyt (commented):
@fioriandrea Thanks for this, and sorry for the long delay. This definitely seems better than before, but I must confess it's difficult to get my head around ... an expression with both lower and higher precedence than all other expressions?

Two things I want to ask before I consider merging:

1) Is it possible to hit the "expected expression instead of %s" error, and if so, can we add a test case for that?
2) Given that it's a recursive parser, but we're only storing one `pendingGetlineLeft` value, is it possible that with a nested expression we'd overwrite the `pendingGetlineLeft` field? I tried a few combinations but couldn't get this to happen, so maybe it can't happen.

fioriandrea:
I will look more closely at the code tomorrow, but for now I can say this:

1. Two possible (incorrect) programs that trigger the condition are: `BEGIN { | }` and `BEGIN { 1 + 1 - | }`. I can add these two programs in the tests
2. I don't think this can happen. In order for pendingGetlineLeft to be overwritten
there would have to be a situation in which pendingGetlineLeft is set and, while
it's set, someone calls Parser.getLine(). But pendingGetlineLeft is set only
*after* we are sure that there is a pipe immediately after. When there is a pipe
immediately after, we know that we will land straight on the PIPE case in the
Parser.primary function, where pendingGetlineLeft is immediately unset anyway.
So there isn't any opportunity for pendingGetlineLeft to be overwritten.

fioriandrea:
I made some modifications to my code in the hopes that the intent is more clear.

First of all, I removed the if condition with the "expected expression instead of %s" error, as the situation can be handled in the default branch of the switch statement in `primary`. In order to do this, I moved the code I had before in its own if at the top of the `primary` function. The intent of having `pendingGetlineLeft` looks more clear with this changes (at least to me). Nevertheless, I added the tests for `BEGIN { | }` and `BEGIN { 1 + 1 - | }`.

For the second point, I think the reasoning I did in my last comment was sound. We parse an arbitrary expression with precedence <= than `cond`, then, if we encounter a PIPE, we know we are inside a getline expression, so we save the lhs in a variable, then we go straight to `primary`, where we finish constructing the rhs of getline and where we unset `pendingGetlineLeft`, leaving no chance of overwriting.

I also thought of putting `p.expect(GETLINE)` inside `if p.tok == PIPE {` in the `getline` function. If we want, we can do that. It doesn't change the behavior and it might make things more clear.

parser/parser.go
fioriandrea: Two possible (incorrect) programs that trigger the condition are: `BEGIN { | }` and `BEGIN { 1 + 1 - | }`

benhoyt:
That looks a bit nicer, and thanks for the additional test cases. Really appreciate you working through this, @fioriandrea! Merging now.

Issue #221, opened 2024-02-26
parser: Discriminate between constant and dynamic regexps


ypdn:
I've been using the `github.com/benhoyt/goawk/parser` package to format my awk scripts, but there is a small issue: after parsing and printing a program (with `Program.String`) "constant" regexps become "dynamic" regexps.

For example:

`{sub(/regexp/, "", $0)}` becomes:
```
{
    sub("regexp", "", $0)
}
```

Arguably it doesn't matter but it would be nice if it didn't change the regexps. Thanks for your work.

benhoyt:
Thanks for the report. `Program.String()` isn't actually intended to be a proper AWK code formatter, but I'm glad that you're able to use it that way. :-)

I'd be in favour of adding this. Currently the AST doesn't distinguish between `/regexp/` and `"regexp"` for these cases, both are just parsed as `*ast.StrExpr`. There is an `*ast.RegExpr` already, but it's currently only used for a stand-alone regexp, so I'm not sure we can re-use that. Probably need a new `ast.FooExpr` type, or we could keep track of which `StrExpr`s are actually `/regexes/`.

benhoyt:
Would you be interested in coming up with a PR by any chance?

ypdn:
> ... or we could keep track of which StrExprs are actually /regexes/

This seems like the better way to do it, since we're only interested in its String method. I've submitted the PR.

PR #222, opened 2024-02-28
Distinguish between /regexp/ and "regexp" in stringified output

ypdn:
Fixes issue #221

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -292,12 +292,17 @@ func (e *NumExpr) String() string {
 	}
 }
 
-// StrExpr is a literal string like "foo".
+// StrExpr is a literal string like "foo" or a regexp constant like /foo/.
 type StrExpr struct {
-	Value string
+	Value  string
+	Regexp bool
 }
 
 func (e *StrExpr) String() string {
+	if e.Regexp {
+		r := &RegExpr{e.Value}
+		return r.String()
+	}
 	return strconv.Quote(e.Value)
 }
 
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -300,8 +300,7 @@ type StrExpr struct {
 
 func (e *StrExpr) String() string {
 	if e.Regexp {
-		r := &RegExpr{e.Value}
-		return r.String()
+		return formatRegex(e.Value)
 	}
 	return strconv.Quote(e.Value)
 }
@@ -313,8 +312,7 @@ type RegExpr struct {
 }
 
 func (e *RegExpr) String() string {
-	escaped := strings.Replace(e.Regex, "/", `\/`, -1)
-	return "/" + escaped + "/"
+	return formatRegex(e.Regex)
 }
 
 // VarExpr is a variable reference (special var, global, or local).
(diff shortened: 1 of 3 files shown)


internal/ast/ast.go
```diff
@@ -292,12 +292,17 @@ func (e *NumExpr) String() string {
...
 type StrExpr struct {
-	Value string
+	Value  string
+	Regexp bool
 }
 
 func (e *StrExpr) String() string {
+	if e.Regexp {
+		r := &RegExpr{e.Value}
+		return r.String()
```
benhoyt, line 304: Nit suggestion (untested, but I think it should work):
```suggestion
		return RegExpr.String(e.Value)
```

internal/ast/ast.go
```diff
@@ -292,12 +292,17 @@ func (e *NumExpr) String() string {
 	}
 }
 
-// StrExpr is a literal string like "foo".
+// StrExpr is a literal string like "foo" or a regexp constant like /foo/.
 type StrExpr struct {
-	Value string
+	Value  string
+	Regexp bool
```
benhoyt, line 298: Naming nit: can we call this `IsRegexp` for clarity?

benhoyt (changes requested):
Nice -- this is a very simple and tidy fix! I hadn't thought of that approach.

Just a couple of nit suggestions, and can we please add a test? You should be able to just add a line or two to the existing `TestParseAndString` test in `parser_test.go`.

internal/ast/ast.go
ypdn: It won't, because the (*RegExpr).String takes a *RegExpr as an argument, not a string.

benhoyt: Ah yes, you're right.

benhoyt: Building a `RegExpr` object here still feels a little weird -- can we factor out the content of `RegExpr.String` into `formatRegex` and call it from both places?

ypdn: Yeah

internal/ast/ast.go
ypdn: We could, but that's so not-Go

ypdn: Maybe we should document it better?

benhoyt: Looks like your mostly right that it's "not Go", though not entirely, [for example, the `Is*` fields here](https://pkg.go.dev/net#DNSError). However, we have another bool in this package, `IncrExpr.Pre`, that doesn't have the `Is`, so let's go with that.

However, on closer look, can we please call it `Regex bool`? The "regexp" with a 'p' kind of refers to Go's "regexp" package, whereas we I tend to call it "regex" in an AWK context (for example, the `RegExpr.Regex` field itself).

ypdn: Done

benhoyt (approved):
Nice, thanks for the contribution!

ypdn:
@benhoyt Could you create a new tag so that `go get` would download it?

benhoyt:
That's okay: 1) the `ast` package is in `internal` so it can't be used outside of this repo, and 2) Go package compatibility usually doesn't guarantee that will continue working -- see "struct literals" in https://go.dev/doc/go1compat

PR #232, opened 2024-05-10
fix: close file

testwill:


benhoyt (changes requested):
Thanks for the fix. Out of interest, how did you find/notice this?

Note that the "defer close" needs to be before the `isNewFile` block, otherwise if that first write fails the file still won't be closed.

Also, because it's a file we're writing to, we need to handle and return the error from `f.Close()`. I suggest changing the return value to a named return value, `(err error)` and doing something like this:

```go
	defer func() {
		closeErr := f.Close()
		err = errors.Join(err, closeErr)
	}()
```

Is there any way we can test for this? I'm guessing it won't be easy, so probably not worth it.

benhoyt (approved):
You forgot the `(err error)` named return and moving the defer up.

I fixed this and also remembered `errors.Join` is Go 1.20+ (currently `go.mod` is on 1.16) so I removed the use of that, see https://github.com/benhoyt/goawk/pull/232/commits/19626538c015aa058d24d2970579289a5a319f7e

benhoyt:
Thanks again for spotting this!

PR #237, opened 2024-08-08
Alternative fix for #236 that avoids decoding entire string

benhoyt:
This is an alternative fix for https://github.com/benhoyt/goawk/pull/236 that avoids decoding the entire string as `[]rune`, and also updates the printf `%c` tests. I was going to update that PR branch but didn't have write access.


benhoyt:
Closing in favour of #236 now that that's updated.

PR #243, opened 2024-09-18
Add Unicode chars mode: "goawk -c" or Config.Chars=true

benhoyt:
Use chars mode in index(), length(), match(), and substr(). This is based on the work done in https://github.com/benhoyt/goawk/pull/83, but the default is the other way around (default = bytes mode, which is the same as it is now).

Also only use Unicode chars in `printf %c` if chars mode is enabled. This changes the default introduced in https://github.com/benhoyt/goawk/pull/236 / v1.28.0 for that feature.

This won't please everyone (defaults are the same as mawk, but not Gawk). But it's a pragmatic solution that allows people to opt into Unicode support if they want, and if they can handle the O(N) behaviour of those functions.

### DIFF
--- a/goawk.go
+++ b/goawk.go
@@ -57,6 +57,7 @@ const (
   -v var=value      variable assignment (multiple allowed)
 
 Additional GoAWK features:
+  -c                use Unicode chars for index, length, match, substr, and %c
   -E progfile       load program, treat as last option, disable var=value args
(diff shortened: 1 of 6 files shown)


JohnEarnest:
Excellent! With these changes, [lila](https://github.com/JohnEarnest/Decker/tree/main/tools/awk) now passes its full test suite under goawk's default settings.

PR #251, opened 2025-06-29
[std]: fix `system()` for Windows

ZoomRmc:
Now resolves to launching `cmd /C` instead of `sh -c`

This, of course, violates the POSIX, which for `system()` states:

>   Execute the command given by expression in a manner equivalent to the
>   `system()` function defined in the System Interfaces volume of
>   POSIX.1-2017 and return the exit status of the command.

Where the `system` function is explicitly `sh`:

>   The system() function shall behave as if a child process were created
>   using fork(), and the child process invoked the **sh** utility using
>   execl() as follows:
>   ```execl(<shell path>, "sh", "-c", command, (char *)0);```

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -990,11 +990,11 @@ func (p *interp) compileRegex(regex string) (*regexp.Regexp, error) {
 }
 
 func getDefaultShellCommand() []string {
-	executable := "/bin/sh"
+	executable, runcmd := "/bin/sh", "-c"
 	if runtime.GOOS == "windows" {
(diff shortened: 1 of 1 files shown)


benhoyt:
Thanks for the contribution. On Windows, we're not in a POSIX environment, so we have to be flexible with the (POSIX) AWK spec. `sh` is not normally present on Windows, so it's better to use `cmd`. You'll note also that this causes the tests to fail on Windows. Given that, I think I'm going to keep the current behaviour.

PR #252, opened 2025-07-28
[WIP] Cache runes in chars mode

triallax:
WIP implementation of my proposal in https://github.com/benhoyt/goawk/issues/35#issuecomment-3111734804

unsophisticated and needs further testing and benchmarking, but already quite promising: running `gron.awk` on a big JSON file in chars mode goes from taking 12(!) minutes to 4 seconds on my laptop with this branch in its current form.

### DIFF
--- a/interp/functions.go
+++ b/interp/functions.go
@@ -53,13 +53,13 @@ func (p *interp) callNative(index int, args []value) (value, error) {
 		return null(), nil
 	case 1:
 		// Single return value
-		return fromNative(outs[0]), nil
+		return fromNative(outs[0], p.chars), nil
 	case 2:
 		// Two-valued return of (scalar, error)
 		if !outs[1].IsNil() {
 			return null(), outs[1].Interface().(error)
 		}
-		return fromNative(outs[0]), nil
+		return fromNative(outs[0], p.chars), nil
 	default:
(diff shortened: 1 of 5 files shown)


triallax:
@benhoyt gentle bump, no rush but i'd love to know how you think this looks for a prototype

benhoyt:
Hi @triallax, I'm not opposed to the approach itself. However, I guess the main thing I'm worried about is that it significantly slows down normal (bytes-mode) performance, due I suppose to all the additional allocations from the `new([]rune)` when creating each new value. I did a test locally (with `simple.awk` from https://github.com/benhoyt/countwords and `kjvbible_x10.txt`) and in bytes mode it was about twice as slow, or took twice as long, with this change.

We could try to have a switch in the interpreter which avoids the allocation if in bytes mode. The wiring for that might be annoying, but at least we could test performance. Even then I'd want to make sure the if/else in the hot path didn't slow bytes-mode down much -- GoAWK's been fairly focussed on performance, and I'd hate to regress much.

In any case, it'd be good if you could show some performance numbers, and we could come up with a way to avoid the extra `new()` / allocation for every string value.

triallax:
(sorry for the delay, been a lil busy, i promise i'll get around to this very soon)

PR #278, opened 2026-04-06
Ensure split() handles string " " differently from regex / /

benhoyt:
- split(in, a, " ") should split like strings.Fields(), trimming whitespace
- split(in, a, / /) should split on a single space

In general, when `/.../` slashes are used, the content is explictly a regex and should be treated as such.

This is as per the other AWKs. (The POSIX spec is a little unclear on this point, but seems to hint at it.)

### DIFF
--- a/goawk_test.go
+++ b/goawk_test.go
@@ -229,12 +229,11 @@ func sortedLines(data []byte) []byte {
 }
 
 func TestGAWK(t *testing.T) {
-	skip := map[string]bool{ // TODO: fix these (at least the ones that are bugs)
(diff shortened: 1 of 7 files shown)


benhoyt:
Hmmm, behaviour related to this was changed in Gawk 5.3.2 as well. Might need to tweak this. See https://cgit.git.savannah.gnu.org/cgit/gawk.git/commit/?h=gawk-5.3.2&id=12e260e3bfedd72f8a2d5f3570e4ff15ec7f006c

PR #291, opened 2026-06-25
Add OpenFile option to Config

vyskocilm:
This allows user to provide a custom file opener instead of a default `os.OpenFile`, making `goawk` compatible with a newer standard library features like `os.Root` protecting against path traversal attacks or simulating a read-only filesystems.

Attached test cases for `os.Root` and a read-only filesystem.

TL;DR;

My motivation is my opensource project https://codeberg.org/gonix/, which is an user-space implementation of a unix scripting in Go. Somewhere `goawk` matches perfectly. Thanks for spending your time on it :heart: The first application of my userspace scripting is a secure, yet shell scriptable, MCP server for agents https://codeberg.org/gonix/mcp.

At the moment it is secured by the fact most of available commands runs under Web Assembly and a shell layer `mvdan.cc/sh/v3` allows one to define a proper handlers for `os.OpenFile`, `os.Stat` or `os.Readdir`. See

https://codeberg.org/gonix/mcp/src/commit/e5cff85075e6191e6415fa66435a4c73470cc978/internal/tools/shell_execute/tool.go#L159-L163

I kept the `OpenFileFunc` compatible with `os.OpenFile` and `os` package, however an another possibility to consider is a form with a context, mainly because a `sh` module uses it to pass various stuff like a working directory there.

```go
// OpenFileFunc specifies a function used to open files instead of the default
// [os.OpenFile]. ctx is a context passed via [Interp.ExecContext], or [context.Background()]
// if [Interp.Exec] has been called.
OpenFileFunc func(ctx context.Context, name string, flag int, perm os.FileMode) (*os.File, error)
```

However it is up to your judgement.

BTW: an another task for an air-gaped `goawk` is to replace `*exec.Command`, but I kept this PR focused on a single topic. Additionally I am not entirely sure how to solve that one :-).


interp/interp.go
```diff
@@ -87,6 +87,7 @@ type interp struct {
 	csvOutput     *bufio.Writer
 	noArgVars     bool
 	splitBuffer   []byte
+	openFileFunc  func(name string, flag int, perm os.FileMode) (*os.File, error)
```
benhoyt: Nit: let's name this field `openFile`. In my experience the "Func" suffix is normally used on the type name (which I think we should add -- see below).

goawk.go
```diff
@@ -355,6 +355,7 @@ argsLoop:
 			"INPUTMODE", inputMode,
 			"OUTPUTMODE", outputMode,
 		},
+		OpenFileFunc: os.OpenFile,
```
benhoyt: `os.OpenFile` is the default, so let's drop this line.

benhoyt: Also, per comment in interp.go about `fs.ErrNotExit`, let's update `errorExit` in this file as follows:

```go
func errorExit(err error) {
	var pathErr *fs.PathError
	if errors.As(err, &pathErr) && errors.Is(err, fs.ErrNotExist) {
	// ...
}
```

interp/interp.go
```diff
@@ -321,6 +322,10 @@ type Config struct {
 	// output. The default is "smart", meaning no translation on Linux/Unix
 	// and CRLF translation on Windows.
 	NewlineOutput NewlineMode
+
+	// OpenFileFunc specifies a function used to open files instead of the default
+	// [os.OpenFile].
+	OpenFileFunc func(name string, flag int, perm os.FileMode) (*os.File, error)
 }
```
benhoyt, line 335: Let's flesh out this doc comment a bit to show what it's used for, and let's add an `OpenFileFunc` type  and use it here (and above in the `interp` struct field). Let's also add documentation about the error contract.

```suggestion

	// OpenFile specifies the function used to open and create files. Set this
	// to an [os.Root.OpenFile] method to limit file access to within a root
	// tree, or to a custom function (for example, to limit the interpreter to
	// read-only file access). The default is [os.OpenFile].
	//
	// A custom OpenFile function must return an error that satisfies
	// errors.Is(err, fs.ErrNotExist) for files that don't exist when flag is
	// os.O_RDONLY.
	OpenFile OpenFileFunc
}

// OpenFileFunc is the type used for setting [Config.OpenFile], and is the type
// of [os.OpenFile].
type OpenFileFunc func(name string, flag int, perm os.FileMode) (*os.File, error)
```

We should also change `if _, ok := err.(*os.PathError); ok {` in vm.go to `if errors.Is(err, fs.ErrNotExist) {` to modernize that.

interp/interp.go
```diff
@@ -565,6 +570,12 @@ func (p *interp) setExecuteConfig(config *Config) error {
 		return fmt.Errorf("invalid newline output mode %d", config.NewlineOutput)
 	}
 
+	if config.OpenFileFunc == nil {
```
benhoyt: Let's move this up to the bottom of the "// Set up I/O structures" block.

interp/io.go
```diff
@@ -203,7 +203,7 @@ func (p *interp) getInputScannerFile(name string) (*bufio.Scanner, error) {
 	if p.noFileReads {
 		return nil, newError("can't read from file due to NoFileReads")
 	}
-	f, err := os.Open(name)
+	f, err := p.openFileFunc(name, os.O_RDONLY, 0)
 	if err != nil {
 		return nil, err // *os.PathError is handled by caller (getline returns -1)
```
benhoyt:
```suggestion
		return nil, err // fs.ErrNotExist is handled by caller (getline returns -1)
```

interp/newexecute_go124_test.go
```diff
@@ -0,0 +1,94 @@
+//go:build go1.24
```
benhoyt: This file isn't really "newexecute" related, so let's rename it, perhaps `openfile_root_test.go`. Also, it's already got a `go:build` constraint for go1.24, so no need for that in the filename as well.

interp/newexecute_go124_test.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+	})
+
+	var output bytes.Buffer
+	status, err := interpreter.Execute(&interp.Config{
+		Stdin:  strings.NewReader(""),
+		Output: &output,
+		OpenFileFunc: func(name string, flag int, perm os.FileMode) (*os.File, error) {
+			return root.OpenFile(name, flag, perm)
+		},
```
benhoyt: Can we simplify by using a method expression?
```suggestion
		OpenFileFunc: root.OpenFile,
```

interp/newexecute_go124_test.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		t.Fatalf("error opening file in root: %v", err)
+	}
+	t.Cleanup(func() {
+		err := f.Close()
+		if err != nil {
+			t.Fatalf("error closing file in root: %v", err)
+		}
+	})
+	data, err := io.ReadAll(f)
```
benhoyt: I don't think we need to use the root or reopen the root here. Can we use `os.ReadFile` and just do this:

```suggestion
    data, err := os.ReadFile(filepath.Join(dir, "output.txt"))
```

interp/newexecute_go124_test.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		t.Fatalf("error reading file in root: %v", err)
+	}
+	const expected = "Hello, GoAWK!"
```
benhoyt: Tests are failing at the moment. I believe this needs to have a newline at the end:
```suggestion
	const expected = "Hello, GoAWK!\n"
```

interp/newexecute_go124_test.go
```diff
@@ -0,0 +1,94 @@
...
+	normalized := normalizeNewlines(string(data))
+	if normalized != expected {
+		t.Fatalf("expected file content %q, got %q", expected, normalized)
+	}
+}
+
+func TestExecuteOpenFileFunc_ReadOnly(t *testing.T) {
```
benhoyt: This test doesn't use Go 1.24 (os.OpenRoot), so should be moved to `interp_test.go` at an appropriate place. Let's also rename to:
```suggestion
func TestOpenFileCustom(t *testing.T) {
```

interp/newexecute_go124_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/benhoyt/goawk/interp"
+)
+
+func TestExecuteOpenFileFunc(t *testing.T) {
```
benhoyt: Nit: let's rename this:
```suggestion
func TestOpenFileRoot(t *testing.T) {
```

interp/newexecute_go124_test.go
```diff
@@ -0,0 +1,94 @@
...
+func TestExecuteOpenFileFunc_ReadOnly(t *testing.T) {
+	source := `BEGIN { print "Hello, GoAWK!" > "output.txt" }`
+	interpreter := newInterp(t, source)
+
+	var output bytes.Buffer
+	status, err := interpreter.Execute(&interp.Config{
+		Stdin:  strings.NewReader(""),
+		Output: &output,
+		OpenFileFunc: func(name string, _ int, _ os.FileMode) (*os.File, error) {
+			return nil, fmt.Errorf("can't open %s for writing: read only filesystem", name)
```
benhoyt: I don't think this tests quite enough. Can we add to this test (or add another one if simpler) that shows a custom OpenFile func that works for reading files but returns an error for writing?

interp/newexecute_go124_test.go
```diff
@@ -0,0 +1,94 @@
...
+		},
+	})
+
+	const expectedErr = `output redirection error: can't open output.txt for writing: read only filesystem`
```
benhoyt: Can we instead assert the error *contains* just the part we're generating here, for example `strings.Contains(err.Error(), "can't open output.txt for writing: read only filesystem")`?

benhoyt (changes requested):
Thanks for this. I'm interested in this change and would like to include it. I think what you have is close, but I've added a few comments about the doc comment, and the error contract, and a few comments on the tests.

interp/interp.go
vyskocilm: Hi @benhoyt thanks for a review.

renamed in 626e8f5

goawk.go
vyskocilm: dropped, modernized the errorExit in goawk.go and the same pattern in interp/vm.go in 626e8f5

interp/interp.go
vyskocilm: done in 626e8f5

interp/newexecute_go124_test.go
vyskocilm: There are three test cases for this custom `OpenFile`

```go
	openFile := func(name string, flags int, mode os.FileMode) (*os.File, error) {
		accMode := flags & (os.O_RDONLY | os.O_WRONLY | os.O_RDWR)
		if accMode == os.O_WRONLY || accMode == os.O_RDWR {
			return nil, fmt.Errorf("can't open %s for writing: read only filesystem", name)
		}
		f, err := os.OpenFile(name, flags, mode)
		return f, err
	}
```

 - cannot write
 - can read
 - getline not found

For the last one I was expecting the status to be -1 and not 0, so maybe I broke some assumption? But the modernized error handling looks correct.

interp/newexecute_go124_test.go
vyskocilm: done in https://github.com/benhoyt/goawk/commit/626e8f5125d625178cac908c01a628e5a5beb937

interp/newexecute_go124_test.go
vyskocilm: Hi @benhoyt,

sure - I never liked the file name: done in dec5590

interp/newexecute_go124_test.go
vyskocilm: Hi @benhoyt,

makes a sense, the test itself can use os.OpenFile. I got more paranoid than needed :-). Added the subtest for a path traversal case, so testing against os.Root triggers the error condition too. See https://github.com/benhoyt/goawk/commit/dec5590555c9a9cbaae9a6ed0b6bc3398895b990

benhoyt:
Thanks for the updates. I'd like to make a few cleanups to the test. To save back and forth, can you please check "Allow edits by maintainers" on your PR branch and I'll push a few changes and then merge? Alternatively I can create a new PR, but I'd like your ownership preserved.

vyskocilm:
> Thanks for the updates. I'd like to make a few cleanups to the test. To save back and forth, can you please check "Allow edits by maintainers" on your PR branch and I'll push a few changes and then merge? Alternatively I can create a new PR, but I'd like your ownership preserved.

Hi @benhoyt thanks for taking a time with a review and improving this PR. Really appreciated.

Allow edit from maintainers checkbox is enabled, so this should work. I am fine with you merging it.

image

benhoyt:
That's odd, when I push I get this:

```
Enumerating objects: 15, done.
Counting objects: 100% (15/15), done.
Delta compression using up to 8 threads
Compressing objects: 100% (10/10), done.
Writing objects: 100% (10/10), 2.30 KiB | 2.30 MiB/s, done.
Total 10 (delta 7), reused 0 (delta 0), pack-reused 0 (from 0)
remote: Resolving deltas: 100% (7/7), completed with 5 local objects.        
To https://github.com/vyskocilm/goawk.git
 ! [remote rejected] vyskocilm/master -> vyskocilm/master (permission denied)
error: failed to push some refs to 'https://github.com/vyskocilm/goawk.git'
```

I'll try again in my morning, otherwise I'll push to a new branch.

vyskocilm:
> That's odd, when I push I get this:
> 
> I'll try again in my morning, otherwise I'll push to a new branch.

This is odd indeed - I invited you as a collaborator inside my fork https://github.com/vyskocilm/goawk/invitations - hope this help.

benhoyt:
Closing in favour of #294 -- after being a collaborator I could push, but for some strange reason the commits didn't show up here.

PR #293, opened 2026-06-29
Add Command option to Config allowing to change the command execution

vyskocilm:
Hi @benhoyt,

This is a follow-up of https://github.com/benhoyt/goawk/pull/291 - given the fact I'll be on vacation next two weeks, I decided to open this one as a draft now. The only interesting part here is e2b7a416a57b3e2c60bbe24812ef514927fdb3c9 

A replacement of `*exec.Cmd` for GoAWK. This may be bit more esoteric, than the previous one. However the context is the same. I build a `shell` compatible environment, which (can) run from userspace running various Go routines instead of binaries via the kernel.

I believe that I found a nice approach for goawk - the `interp.Cmd` interface is a _minimal_ change to the existing code, which makes it easy to review and won't bring bugs into the codebase. Additionally this PR adds (quite complicated, I must say) mock implementation of the interface, which tests the command execution of goawk without any dependency on underlying OS.

BTW As a part of an "integration" testing I managed to implement it in the goawk wrapper

 - WIP PR: https://codeberg.org/gonix/utils/pulls/12

Additionally I have a test connecting the `codeberg.org/gonix/sh` and `mvdan.cc/sh/v3` with goawk with this interface. This connects all the dots making following awk script working without a `/bin/sh`. I'll update this PR with a link to a proper PR once I'll manage to create one.

```awk
BEGIN {
      while (("gonix/seq 3") | getline n)
          sum += n
      print sum
  }`
```

Thanks a lot for your review!


benhoyt:
I will take a closer look at this, but just to set expectations, I probably won't merge this. I'd rather not introduce this level of complexity for a niche use case.

vyskocilm:
> I will take a closer look at this, but just to set expectations, I probably won't merge this. I'd rather not introduce this level of complexity for a niche use case.

Hi @benhoyt yeah, understandable - my project is very specific. But let me push back a bit :pray: . I think the change itself is very minimal and keep the original code as much as possible. In fact the ugliest part was the mock and unit tests, which can be removed. Additionally the mvdan.cc/sh/v3, which is in a similar Go reimplementation of Unix sphere, allows one to replace `*exec.Cmd`.

Anyway I'll keep the final decision to you - if you do not want this in your project, or if you're not interested in reviewing this, feel free to close this PR.

benhoyt:
I've thought further about this, and I'm going to decline (but thanks for the contribution). If we do do this at some stage, I want to think carefully about the interface (which is new public API surface, so we're locked in forever), and probably simplify somewhat and avoid GoAWK logic like what's in WaitExitCode. So it's not a "no never" but a "not now, and not in this form". In the meantime go ahead and fork!

PR #296, opened 2026-07-28
Replace Config.OpenFile with more general fs.FS-based approach

benhoyt:
In #294 we added Config.OpenFile, but it returned a concrete *os.File, which makes it a lot less flexible than it needs to be. As this version hasn't been released yet, let's see if we can improve that.

I'd like to fit in with io/fs.FS if possible, so that's what I've tried to do: if you provide a bare fs.FS for Config.FileSystem it allows read-only file access. If the type also implements WriteFS (adding Create and Append methods) it allow read and write access.

This is a breaking change, but (thankfully) I haven't released the Config.OpenFile support in a versioned release yet, so I'm okay with this.

### DIFF
--- a/goawk_test.go
+++ b/goawk_test.go
@@ -113,7 +113,7 @@ func TestAWK(t *testing.T) {
 				expected = sortedLines(expected)
 			}
 			if writeAWK {
-				err := os.WriteFile(outputPath, expected, 0644)
+				err := os.WriteFile(outputPath, expected, 0o644)
 				if err != nil {
 					t.Fatalf("error writing awk output: %v", err)
 				}
@@ -139,7 +139,7 @@ func TestAWK(t *testing.T) {
 				output = sortedLines(output)
 			}
 			if writeGoAWK {
-				err := os.WriteFile(outputPath, output, 0644)
+				err := os.WriteFile(outputPath, output, 0o644)
 				if err != nil {
 					t.Fatalf("error writing goawk output: %v", err)
 				}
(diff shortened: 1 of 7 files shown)


vyskocilm (approved):
Hi @benhoyt I think this is a neat idea. The *os.File is unflexible and would limit goawk's usability. Using the `io/fs` interfaces is probably the best approach you can choose.

I like the `Create` and `Append` naming hiding the ugly `os.O_` flags behind a nice interface.

benhoyt:
@vyskocilm Care to review this code change? I'll wait a couple of days before merging.

vyskocilm:
@benhoyt I found a little time, so rebased my goawk fork on top this and implemented by gonix awk wrapper. Note this PR will fail in CI as I use `go.work` internally and don't want to push the `goawk` fork itself. However you can check how this change will be consumed and that its pretty straighforward.

https://codeberg.org/gonix/utils/pulls/15

I have updated the mcp too - however I accidentally included this in a pile of other refactoring which I do not want to publish even as a WIP. Sorry :-).

My feedback is **a go** - the `fs.FS` vs `WriteFS` is clear.

There may be a helper like `func CastWriteFS(fs fs.FS) (WriteFS, error)`, which will ensure all parts and wrappers returns the same error for read only FS. But that's only my personal taste. Feel free to reject.

benhoyt:
Thanks for the review. I think I'll leave out `CastWriteFS` and merge this as is.

PR #297, opened 2026-08-20
Default bare `%g`/`%G` to 6 significant digits like C/POSIX awk

jaideeppyne:
## What

`printf` with a bare `%g`/`%G` (no explicit precision) diverges from C/POSIX awk:

```
printf "%g", 12345.678    → goawk: 12345.678    gawk/onetrueawk/mawk/C: 12345.7
printf "%g", 2.0/3.0      → goawk: 0.666666666666...   others: 0.666667
```

## Why

goawk hands format verbs to Go's `fmt.Sprintf`. Go's `%g`/`%G` **without an explicit precision** emit the *shortest round-trippable* representation, whereas C/POSIX `%g` defaults to **6 significant digits**. In `parseFmtTypes` the `g`/`G` case sets the argument type to `'f'` but never injects the default precision, so the bare verb is passed through to Go's shortest-form behaviour.

`%e`/`%f` are unaffected (Go and C both default to 6), and explicit precision (`%.3g`, `%.*g`) already works — only the bare verb was wrong. The `print`/`OFMT` path is also unaffected since it hard-codes `%.6g`.

## How

In `parseFmtTypes`, inject a default precision of `.6` for a bare `%g`/`%G` (i.e. when no `.` precision appeared in the spec). The output format is now built incrementally instead of mutating a copy of the format string in place, because inserting `.6` shifts later indices.

Cross-checked against `gawk` — bare `%g`/`%G` now match, and `%.3g`, `%.*g`, `%#g`, `%+g`, `%-10g`, `%*g`, `%e`, `%f`, `%a`, and the integer/char verbs are all unchanged:

| format | before | after / gawk |
|---|---|---|
| `%g`, 12345.678 | `12345.678` | `12345.7` |
| `%G`, 12345.678 | `12345.678` | `12345.7` |
| `%g`, 0.666666666666 | `0.666666666666` | `0.666667` |
| `%.3g`, 12345.678 | `1.23e+04` | `1.23e+04` |
| `%#g`, 100 | `100.000` | `100.000` |

## Tests

Added `%g`/`%G` cases to the `interpTests` table in `interp/interp_test.go`. They **fail on `master`** (`got "12345.678"`) and pass with the fix. The interp harness also cross-checks each case against the installed `awk`/`gawk`, and `go test ./...` passes across the module.

---

*Disclosure: this change was prepared with AI assistance and reviewed/verified by me before submission.*

### DIFF
--- a/interp/functions.go
+++ b/interp/functions.go
@@ -339,55 +339,76 @@ func (p *interp) parseFmtTypes(s string) (format string, types []byte, err error
 		return item.format, item.types, nil
 	}
 
-	out := []byte(s)
+	// Build the output incrementally (rather than mutating a copy of s in place)
+	// so we can inject a default precision for bare %g/%G, which shifts indices.
+	out := make([]byte, 0, len(s)+2)
 	for i := 0; i < len(s); i++ {
-		if s[i] == '%' {
-			i++
-			if i >= len(s) {
-				return "", nil, errors.New("expected type specifier after %")
-			}
-			if s[i] == '%' {
-				continue
-			}
-			for i < len(s) && strings.IndexByte(" .-+*#0123456789", s[i]) >= 0 {
-				if s[i] == '*' {
-					types = append(types, 'd')
-				}
-				i++
-			}
-			if i >= len(s) {
-				return "", nil, errors.New("expected type specifier after %")
-			}
-			var t byte
-			switch s[i] {
-			case 's':
-				t = 's'
-			case 'd':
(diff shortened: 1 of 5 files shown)


benhoyt:
Thanks for this! This looks like a legitimate discrepancy from all the other AWKs, so I'm happy to fix it. I'll push a few updates to the PR branch (including the similar fix for OFMT and CONVFMT) and merge once I'm happy.

PR #300, opened 2026-08-26
interp: print large integral numbers in full instead of scientific notation

jaideeppyne:
goawk formatted a number as an integer only when it fit in an `int64`, so integral values beyond that range fell through to the `%g` float format (and OFMT/CONVFMT):

```
$ goawk 'BEGIN { print 1e20 }'
1e+20
```

gawk and the one true awk both print such values in full:

```
100000000000000000000
```

The `int64` conversion could also overflow: `print 2^63` printed `9223372036854775807` (int64 max) instead of the correct `9223372036854775808`.

Per POSIX, integral values are converted without reference to OFMT/CONVFMT. This detects integral values with `math.Trunc` and, when they're outside the int64 range, formats them with `strconv.FormatFloat(n, 'f', 0, 64)` — the exact decimal digits, no exponent — matching gawk and onetrueawk. The int64 range keeps the fast `FormatInt` path. This also fixes cases like `OFMT="%.2f"; print 1e19`.

Added interpreter tests (validated against gawk) and updated one existing big-number test that now matches gawk exactly (previously skipped for gawk). `go test ./...`, gofmt, and go vet all pass.

Note: open PR #299 also touches this function; if it merges first this needs a trivial rebase (the same `case` moves into the new `numToStr` helper).

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -305,10 +305,16 @@ BEGIN {
 	{`BEGIN { RS="x" } /^a.*c$/`, "a\nb\nc", "a\nb\nc\n", "", ""},
 	{`BEGIN { print "-12"+0, "+12"+0, " \t\r\n7foo"+0, ".5"+0, "5."+0, "+."+0 }`, "", "-12 12 7 0.5 5 0\n", "", ""},
 	{`BEGIN { print "1e3"+0, "1.2e-1"+0, "1e+1"+0, "1e"+0, "1e+"+0 }`, "", "1000 0.12 10 1 1\n", "", ""},
-	{`BEGIN { print -(11102200000000000000000000000000000000 1040000) }  # !gawk - gawk supports big numbers`,
-		"", "-inf\n", "", ""},
+	{`BEGIN { print -(11102200000000000000000000000000000000 1040000) }`,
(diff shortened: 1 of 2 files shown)


benhoyt:
Should this also apply to the other places we use the float64(int64(x)) pattern?

jaideeppyne:
Yes. The remaining `float64(int64(x))` sites were the same integer check, so they overflowed the same way once the value left the int64 range.

Addressed in 898b18f81c4fcd7a7ece0a6553c7d0b425daa25a:

- `ast.FormatIntegral` now uses `math.Trunc` (with `FormatFloat('f', 0, 64)` outside int64) and is shared by `NumExpr.String()`, the integer-constant array-index compile, and the disassembler.
- `int()` was the other `float64(int64(x))` use: that is truncation, not formatting, but it had the same overflow (`int(1e20)` became `9223372036854775807`). It now uses `math.Trunc` so large integrals stay exact, matching POSIX / gawk. NaN still maps to 0, as the previous int64 conversion did.

I left the `int64(v.num())` conversions in `interp/functions.go` alone — those are FFI into Go integer types, not AWK integer detection.

Added tests for `int(1e20)` / `int(2^63)` / `int(-1e20)` and `a[1e20]`.

